PR #4, opened 2023-05-27
Update topologies and add IQM5q

stavros11:
Updates `topology` in all runcards to use the adjacency matrix as is required after the merge of qiboteam/qibolab#387.

Also adds the runcard and platform for IQM-5q since now the Zurich driver is on qibolab main.


AleCandido (approved):
Everything fine, having all runcards is also very useful for creating the database (I'm planning to do it soon, at least a sketch).

The only weird thing is that you don't have all the newlines at the end of the files:
image

stavros11:
> The only weird thing is that you don't have all the newlines at the end of the files: image

Good point, thanks. It's due to not spending time configuring my editor. I added the pre-commit config we have in the other projects to avoid such issues. In principle we could also set up the bot, but I am not sure if I can (and how) to do it.

scarrazza:
Bot enabled, thanks.

iqm5q.py
```diff
@@ -0,0 +1,157 @@
+import pathlib
+
```
Jacfomg, line 2: import itertools
import yaml

AleCandido: If you plan to use them, just commit it (together with their usage).

In more details: since we're not specifying dependencies for these files, it would be slightly more complicate
- `itertools` is standard library, so this one is fully uncontroversial (if it's useful, use it)
- `pyyaml` is not standard library, but in practice is a dependency of Qibolab, and these files are loaded by Qibolab

However, in principle we might want to drop the `pyyaml` dependency in Qibolab, at some point.
So, if you can live without, try avoiding `import yaml`.

Jacfomg: They are already being used on the file. For `itertools` we have the part where we set the channels and qubits and `pyyaml` is for the Zurich config. These last I will try to look for an alternative to what I was using now then.

Jacfomg: @stavros11 or @AleCandido  I can't commit here the changes to `iqm5q.py` because I don't have the permision to push

AleCandido: Not even on the branch?

stavros11: @scarrazza can you please give permission to the tii team to push to this repository? They are the ones that will be using it mostly.

andrea-pasquale:
I've uploaded here the new runcard for `tii1q_b1`

iqm5q.py
AleCandido: Now as an issue #6

scarrazza: Fixed, please retry now.

Jacfomg (approved):

PR #5, opened 2023-05-29
Simplify qw25q

AleCandido:
If you wish, this is just a personal game: I want to see how simple we can trim down the most complex platform that we have.

My hope is that, playing this game, we might design useful abstractions, that can be lifted to Qibolab as building blocks (as it happened for the `ChannelMap`, that looks quite clean to me, despite being just a thin wrapper around a `dict`).

### DIFF
--- a/qw25q.py
+++ b/qw25q.py
@@ -172,29 +172,13 @@ def create(runcard=RUNCARD):
                             qubits[q].drive_frequency + 200e6
                         )
 
-    for q in [
-        "A3",
-        "A5",
-        "A6",
-        "B4",
-        "B5",
-        "C2",
-        "C3",
-        "C5",
-    ]:  # Qubits with LO around 7e9
-        qubits[q].readout = channels[wiring["readout"][q[0]][0]]
-        qubits[q].feedback = channels[wiring["feedback"][q[0]][0]]
-    for q in [
-        "A1",
-        "A2",
-        "A4",
-        "B1",
-        "B2",
-        "B3",
-        "C1",
-        "C4",
-    ]:  # Qubits with LO around 7.5e9
-        qubits[q].readout = channels[wiring["readout"][q[0]][1]]
-        qubits[q].feedback = channels[wiring["feedback"][q[0]][1]]
+    qubits_groups = (
+        ["A3", "A5", "A6", "B4", "B5", "C2", "C3", "C5"],  # Qubits with LO around 7e9
+        ["A1", "A2", "A4", "B1", "B2", "B3", "C1", "C4"],  # Qubits with LO around 7.5e9
+    )
+    for idx, gr in enumerate(qubits_groups):
+        for q in gr:
+            qubits[q].readout = channels[wiring["readout"][q[0]][idx]]
+            qubits[q].feedback = channels[wiring["feedback"][q[0]][idx]]
 
     return platform
--- a/qw25q.py
+++ b/qw25q.py
@@ -116,23 +116,19 @@ def create(runcard=RUNCARD):
         "C": [f"era_0{i}" for i in range(3, 8)],
         # "D": ["era_08"] + 2 * ["LO_01"],
     }
+
     # Configure local oscillator's frequency and power
+    lo_settings = {
+        "LO_01": (6.15e9, 21),
+        "LO_04": (7.1e9, 23),
+        "LO_03": (7.8e9, 23),
+        "LO_05": (5.37e9, 18),
+        "LO_06": (6.2e9, 21),
+    }
     for lo in local_oscillators:
-        if lo.name == "LO_01":
-            lo.frequency = 6.15e9
-            lo.power = 21
-        elif lo.name == "LO_04":
-            lo.frequency = 7.1e9
-            lo.power = 23
-        elif lo.name == "LO_03":
-            lo.frequency = 7.8e9
-            lo.power = 23
-        elif lo.name == "LO_05":
-            lo.frequency = 5.37e9
-            lo.power = 18
-        elif lo.name == "LO_06":
-            lo.frequency = 6.2e9
-            lo.power = 21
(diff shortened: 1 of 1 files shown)


AleCandido:
@DavidSarlle in this way, we could also move towards a more declarative approach, in the sense that if we have repeated patterns, we just need to specify the parameters of these patterns, passing them to suitable functions.

If in the end everything will be covered by suitable building blocks, then we can eventually collect back the parameters, and even avoid the Python file at all.
But for the time being, it is much simpler to use the flexibility of a GP language, rather than creating our own.

DavidSarlle:
Thank you, @AleCandido , for your comments. From my point of view, it is clear that all setups in different laboratories have repeated patterns. The difference lies in the instrumentation used. For this reason, in my opinion, the Platform could be composed (as it already does with the new approach, but with the added capability of interpreting external configuration data, whether it's in YAML, a database, or any other system) of an instrument controller (a generic controller capable of controlling any supported instrument), the class that handles connections (channel/channelMap), and the qubits class. This is the most dynamic approach that facilitates the inclusion of new labs/users. It also ensures that adding new instruments does not require modifying any part of the qibolab or qibocal code, thereby simplifying the amount of code to maintain (there is now a new repository with one file per setup and one per runcard).

Additionally, it's important to consider that users who are using qibolab for the first time in their laboratory are experimental teams or chip manufacturers who generally do not have deep knowledge of programming or system administration. That's why avoiding the need to modify/create Python files is a great advantage for them. For example, the Boulos team, who do not know how to program, feels much more comfortable with the idea of configuring the lab setup by filling out a config template (similar to the qw5q_qblox_runcard). Their ideal scenario would even be to perform this configuration through a graphical interface. That would be great for them.

Regarding the other decisions (not mixing physical parameters and setup configuration in the same runcard, not having code linked to a specific laboratory in the repo...), I completely agree.

For all these reasons and with the premise of allowing users to focus even more easily on high-level concepts, I still think this approach would be more appropriate. But if others see it differently, there is no problem on my side, of course!

So once again, thank you for the explanations.

stavros11:
Something possibly related to this discussion that came up while trying to refactor the `MultiqubitPlatform`, is the `instruments` section in the qblox runcards. Since we are moving to the declarative approach where we have a `create` method for each platform, instruments will be initialized there. However, the instrument settings https://github.com/qiboteam/qibolab_platforms_qrc/blob/55f80d03525ca3497dcb4cc4fb578156503384e5/qw5q_gold.yml#L78

should be stored in the runcard/database, so that we can track potential changes during calibration. Therefore we need a mechanism to load these settings and set the instruments appropriately. This may look different depending on what instruments are used. The `Channel` described in https://github.com/qiboteam/qibolab/issues/429#issuecomment-1568552130 could be responsible for that, however I do not think that it is a good place to store such parameters, because there are instruments (eg. local oscillators) that are shared between different channels, so storing instrument settings in the channels may lead to repetition and bugs.

AleCandido:
Since they are `Instrument`s parameters, the `AbstractInstrument` itself (that btw I'd propose to rename just `Instrument`) could define the API for this:

```py
@dataclass
class InstrumentConfigs:
    """To be optionally subclassed by various instruments.

    It should not contain any method, that should be defined in :cls:`Instrument`,
    but rather just declare the tunable parameters.

    """

@dataclass
class Instrument:
    configs: InstrumentConfigs

    def setup(self):
        self.load(self.configs)

# Just a sketchy example, we should have a look at actual ones
```

AleCandido:
> For this reason, in my opinion, the Platform could be composed (as it already does with the new approach, but with the added capability of interpreting external configuration data, whether it's in YAML, a database, or any other system) of an instrument controller (a generic controller capable of controlling any supported instrument), the class that handles connections (channel/channelMap), and the qubits class. This is the most dynamic approach that facilitates the inclusion of new labs/users. It also ensures that adding new instruments does not require modifying any part of the qibolab or qibocal code, thereby simplifying the amount of code to maintain.

@DavidSarlle about this I quite agree with you. My present opinion is that getting there in a single shot is complex, because we need some time to abstract differences and create sensible APIs to handle all of them.
Moreover, I perfectly agree that the parameters should be stored separately and declaratively, and for this reason I'm working on the DB.
However, some parts of the setup are really working bad declaratively, as the channels: if you have 25 qubits then you might have 25*4 channels. With more qubits they would grow even more, and we'd all start programming our editors to edit those files. Alternatively we would create a syntax in YAML for making loops (well, if sufficiently restrictive, it might be sensible, but not as the first attempt).

> (there is now a new repository with one file per setup and one per runcard)

Well, runcards were already there, we just moved mainly to separate PRs about runcards from PRs about code (and they will eventually fit in the DB), and the goal of this PR is to trim the setup-file more and more.

> Additionally, it's important to consider that users who are using qibolab for the first time in their laboratory are experimental teams or chip manufacturers who generally do not have deep knowledge of programming or system administration. That's why avoiding the need to modify/create Python files is a great advantage for them. For example, the Boulos team, who do not know how to code, feels much more comfortable with the idea of configuring the lab setup by filling out a config template (similar to the qw5q_qblox_runcard). Their ideal scenario would even be to perform this configuration through a graphical interface. That would be great for them.

As said above: this is a lot of work, because if you want to scale, you should allow to algorithmically set channels (e.g.) with a graphical tool.
I'm not saying we will never want this, but we just don't have the manpower to do it now. So let's allow for some flexibility, at no expense.
To support Boulos, it's easier to help him write a tiny `create_` function, rather than creating a specific syntax in YAML.

However, if we're good enough to extract patterns from the `create_` such that they can fit in a declarative format, and still scale, so much the better.
My starting point would be to just simplify `create_`, but if you prefer following a different route I'm open to discussion and proposals.

scarrazza:
Shall we close this?

DavidSarlle:
yes, we can close it @scarrazza

PR #8, opened 2023-05-30
qw5q_gold 20230530 runcard update

aorgazf:


DavidSarlle:
@aorgazf teh topoly defined in this runcard will not work with qibolab main. You should modify it (see the current runcard on main)

scarrazza:
Closing in favor of #34

PR #10, opened 2023-06-01
Rfsoc4x2

rodolfocarobene:
create function update, related to [rfsoc new driver](https://github.com/qiboteam/qibolab/pull/373)

### DIFF
--- a/tii1q_b1.py
+++ b/tii1q_b1.py
@@ -1,17 +1,20 @@
 import pathlib
 
 from qibolab.channels import Channel, ChannelMap
-from qibolab.instruments.rfsoc import TII_RFSOC4x2
+from qibolab.instruments.rfsoc import RFSoC as TII_RFSOC4x2
 from qibolab.instruments.rohde_schwarz import SGS100A as LocalOscillator
 from qibolab.platform import Platform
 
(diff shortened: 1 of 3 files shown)


tii1q_b1.py
```diff
@@ -1,17 +1,20 @@
 import pathlib
 
 from qibolab.channels import Channel, ChannelMap
-from qibolab.instruments.rfsoc import TII_RFSOC4x2
+from qibolab.instruments.rfsoc import RFSoC as TII_RFSOC4x2
```
scarrazza, line 4: For the sake of simplicity, I think we should avoid the `TII_` prefix and just use `RFSoC` directly.

AleCandido:
I assume this is not yet ready, since the files contained are not changing substantially what it is in `main`.
In case, can you make it *draft*?

Otherwise, if I'm wrong, please ask for a review :)

rodolfocarobene:
sorry, I forgot about this PR... it was ready (it's merely an update to reflects the new driver).
However now the qubits will need a re-characterization.

JavierSerranoGarcia (approved):

AleCandido (approved):

PR #13, opened 2023-06-07
Cleaner runcard from the 2q interaction for iqm5q

Jacfomg:
Cleaner runcard from the 2q interaction for iqm5q

### DIFF
--- a/iqm5q.yml
+++ b/iqm5q.yml
@@ -4,7 +4,7 @@ description: IQM 5-qubit device, controlled with Zurich Instruments.
 settings:
     nshots: 4096
     sampling_rate: 2.e+9
-    relaxation_time: 300_000
+    relaxation_time: 100_000
     time_of_flight: 280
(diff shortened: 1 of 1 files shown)


iqm5q.yml
```diff
@@ -19,233 +19,188 @@ native_gates:
...
-            mean_exc_states: -0.8083+2.3929j
+            mean_gnd_states: -1.8517+0.6431j
+            mean_exc_states: -0.8696+1.1473j
             # parameters for single shot classification
-            threshold: -0.0593
-            iq_angle: -0.667
-            # alpha: 208 MHz
+            threshold: -0.6085
+            iq_angle: -0.474
+            alpha: 211_604_296
```
stavros11, line 166: Is this `alpha` properly loaded in the platform/qubits? Because if it is not a field of the `Qubit` dataclass in qibolab, I am not sure if it will be properly loaded.

PR #14, opened 2023-06-12
Zurich ports

Jacfomg:
From https://github.com/qiboteam/qibolab/pull/486 ports update for Zurich Instruments

### DIFF
--- a/iqm5q.py
+++ b/iqm5q.py
@@ -2,7 +2,7 @@ import itertools
 import pathlib
 
 from qibolab import Platform
-from qibolab.channels import ChannelMap
+from qibolab.channels import Channel, ChannelMap
 from qibolab.instruments.oscillator import LocalOscillator
 from qibolab.instruments.zhinst import Zurich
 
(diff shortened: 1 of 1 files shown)


stavros11 (approved):
This can be merged now.

Please remove the `filter: {}` from all qubits in the characterization section of the runcard yaml. It will probably cause an error because I removed it from the `Qubit` object. The reason is that I think it is a QM specific thing so it doesn't need to be a global `Qubit` parameter. If more instruments need it we can put it back.

rodolfocarobene (approved):

Jacfomg:
> This can be merged now.
> 
> Please remove the `filter: {}` from all qubits in the characterization section of the runcard yaml. It will probably cause an error because I removed it from the `Qubit` object. The reason is that I think it is a QM specific thing so it doesn't need to be a global `Qubit` parameter. If more instruments need it we can put it back.

It should be a global thing as far as I know. The problem is we never got them as cryoscope was not running properly.

PR #23, opened 2023-07-10
Add usage instructions on README

stavros11:


README.md
```diff
@@ -1 +1,20 @@
 # qibolab_platforms_qrc
+
```
andrea-pasquale, line 2:
```suggestion
# QRC platforms
```
A better title (?)

andrea-pasquale (approved):
Thanks @stavros11.
I just put a suggestion for the title of the README.

Jacfomg (approved):

rodolfocarobene:
It's good for me. Only one thing: probably is implicit, but I would explicitly write that after the export one has to open a new terminal or source the .bashrc

stavros11:
> It's good for me. Only one thing: probably is implicit, but I would explicitly write that after the export one has to open a new terminal or source the .bashrc

`export` takes effect (only) in the current terminal, so you don't need to open a new one. 
If you put it in `.bashrc` then indeed it will take effect from the next terminal unless you source it or export in the current.

rodolfocarobene (approved):

PR #30, opened 2023-07-14
Change port in qblox

andrea-pasquale:
@Jacfomg discovered that there was a faulty port in qblox.
We replaced it with another one. We tested this branch and it seems to work as expected.


aorgazf (approved):

aorgazf:
Thanks @Jacfomg

PR #32, opened 2023-07-19
Platform for qw25q line B

stavros11:
Following our discussion today, here is the runcard of line B of the qw25q chip that is currently connected to Quantum Machines. 

The issue I have is noise in the low power readout as shown in the following punchout. The same is true for all qubits.

<details>
<summary>Qubit B2 punchout</summary>

image

</details>

Maybe this is because of some parameters set wrong, I am not sure yet.

Also, the drive LOs are not set properly as I have not been able to do qubit spectroscopies yet. Depending on the qubit frequencies, we may need to rewire some LOs or use less.

If anyone wants to give a try, it can be used as platform `qw25qB` from the `qw25q` queue. Here is the qibocal runcard I use:
```yaml
platform: qw25qB

qubits: ["B1", "B2", "B3", "B4", "B5"]

actions:

  #- id: resonator_spectroscopy high power
  #  priority: 0
  #  operation: resonator_spectroscopy
  #  # main: resonator punchout
  #  parameters:
  #    freq_width: 10_000_000
  #    freq_step: 10_000
  #    amplitude: 0.05
  #    power_level: low
  #    nshots: 1024
  #    relaxation_time: 5000

  - id: resonator punchout
    priority: 0
    operation: resonator_punchout
    #main: resonator_spectroscopy low power
    parameters:
      freq_width: 6_000_000
      freq_step: 100_000
      min_amp_factor: 0.001
      max_amp_factor: 2.0
      amplitude: 0.05
      step_amp_factor: 0.01
      nshots: 1024
      relaxation_time: 5000

  # - id: resonator spectroscopy low power
  #   priority: 0
  #   operation: resonator_spectroscopy
  #   #main: qubit_spectroscopy
  #   parameters:
  #     freq_width: 10_000_000
  #     freq_step: 10_000
  #     amplitude: 0.005
  #     power_level: low
  #     nshots: 1024
  #     relaxation_time: 5000

  #- id: resonator flux dependence
  #  priority: 0
  #  operation: resonator_flux
  #  # main: qubit spectroscopy
  #  parameters:
  #    freq_width: 10_000_000
  #    freq_step: 500_000
  #    bias_width: 0.2
  #    bias_step:  0.005
  #    nshots: 1024
  #    relaxation_time: 5000
```


lucamedeir:
So I checked the LO oscillator for the TWPA, 192.168.0.36. It shows 5 GHz 18 dBm. These are the wrong specifications for the TWPA. It should be 6482 MHz 10.2 dBm.

stavros11:
> So I checked the LO for the TWPA, 192.168.0.36. It shows 5 GHz 18 dBm. These are the wrong specifications for the TWPA. It should be 6482 MHz 10.2 dBm. I did notice that in the runcard the settings are correct for the twpa, but when I check the web interface the settings are wrong. Is it possible that another job is changing the value of the same LO6?

Yes, this can happen if another job uses a platform that has the same LO with different settings. However, I just tested it and once I submit a job using the qw25qB platform defined here, the LO settings will be updated to the ones you mentioned (6482 MHz - 10.2 dBm). So I don't think this is the cause of the noise, as it still remains.

stavros11:
Closing this because qw25q will not be available for the time being.

PR #39, opened 2023-08-01
Two qubit gates calibration for qw5q_gold

andrea-pasquale:
I decided to open this PR to keep everyone up to date with the development of two qubit gates for `qw5q_gold`.
This is the new runcard after the power outage.
Currently pair 32 and 12 are "calibrated" . The leakage could be reduced but at least I am able to run both chevron and tune landscape for both pairs.

Chevron 12
image

Tune landscape 12
image

Chevron 32
image

Tune landscape 32
image

We still need to run bell to certify how good our CZ currently is.
After that we will try to run Mermin.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
       - id: check-yaml
       - id: debug-statements
   - repo: https://github.com/psf/black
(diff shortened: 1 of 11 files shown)


andrea-pasquale:
I'm closing this PR since it is outdated.

PR #40, opened 2023-08-02
update after warmup iqm5q0

Jacfomg:
That would be the ones performing best today, not sure if tomorrow will change slighty due to further thermalization.
Also it took at least 3 days after the warmup to reach previous T1 values using the same runcard.


Jacfomg:
I was muted during the meeting, so this should be ready to merge if it performs on the experiments

Jacfomg:
The 2q gate stuff creates a error with iqm on the `native.py`. I will try to investigate why, for now I will just comment it.

Jacfomg:
Good T1 (35s), Assigment fidelity (>.9) and flipping as expected. Still AllXY is weird.

PR #43, opened 2023-08-04
Recalibration after sweetspot shift

andrea-pasquale:
Overnight it seems that the sweetspot shifted in qw5q_gold.
Here is a recalibrated runcard with fidelities similar to those available before.


andrea-pasquale:
Closing this as outdated

PR #49, opened 2023-08-29
Couplers

Jacfomg:
Fixes due to the changes in https://github.com/qiboteam/qibolab/pull/508 when adding coupler support.


iqm5q.py
```diff
@@ -73,7 +79,7 @@ def create(runcard_path=RUNCARD):
         descriptor,
         use_emulation=False,
         time_of_flight=75,
-        smearing=50,  # time_of_flight=280, smearing=100
+        smearing=50,
     )
 
     # Create channel objects and map controllers
```
DavidSarlle, line 85: Sweep labels for readout and feedback channels:

 # feedback
    channels |= Channel(
        "L2-7", port=controller[("device_shfqc", "[QACHANNELS/0/INPUT]")]
    )
    # readout
    channels |= Channel(
        "L3-31", port=controller[("device_shfqc", "[QACHANNELS/0/OUTPUT]")]
    )

iqm5q.yml
DavidSarlle, line 1: Checking th epower levels for RO line set in th eplatform, seems that you are using the latest characterization that I found, at least for q0. If so, please, update q0, q1, q2 attributes and pulses to work porpoerly with the platform config that you are using.

Jacfomg:
@DavidSarlle all those change are already on your PR #54, right ? What is the point on putting them here ?

DavidSarlle:
The point is, if you are testing and checking the physics when using the couplers, you need to add all of them, not only some of them as you did It...

stavros11 (approved):
@Jacfomg in the last push I removed some parameters from iqm5q.yml so that the platform creation tests pass with qiboteam/qibolab#508 and we are ready to merge tomorrow. Let me know you if you actually need any of the removed parameters.

Jacfomg:
> @Jacfomg in the last push I removed some parameters from iqm5q.yml so that the platform creation tests pass with [qiboteam/qibolab#508](https://github.com/qiboteam/qibolab/pull/508) and we are ready to merge tomorrow. Let me know you if you actually need any of the removed parameters.

I think having the fast reset there was convinient, but also as no other instrument cares about it we can remove it.

stavros11:
> I think having the fast reset there was convinient, but also as no other instrument cares about it we can remove it.

If you need it we need to put it in the [`platform.Settings`](https://github.com/qiboteam/qibolab/blob/5ca4e7a294c6c0e7ab2b099bcebcf2e7b8d972a4/src/qibolab/platform.py#L22C7-L22C15) object in qibolab otherwise the platform creation will fail. But I thought it was not working well anyway so maybe we can put it back when implemented properly.

In principle you can still switch it on using the `ExecutionParameters` when executing the sequence, like with `nshots` and `relaxation_time`. It is just that the default value is False and there is no way to change this.

Jacfomg:
I will re run the jobs, they should pass now, right ?

stavros11:
> I will re run the jobs, they should pass now, right ?

Qibolab 0.1.2 seems to be on pypi so they should pass with both stable and main.

Jacfomg:
Yeah, I expected so, but only main did

PR #52, opened 2023-09-19
Test platform creation

### LINKED ISSUE #44: CI
Just as a very cheap sanity check, we should try to execute the `create` functions in the CI, to test that they are actually valid with some version of Qibolab.

Most likely, there is no other sensible check we could do in the CI to test the validity of `create` functions and runcards, because most of the content consist of parameters representing the lab.
But Qibolab compatibility could be tested, and thus we should (at least we'll avoid stale runcards).

About the version choice, there are a few options:
- use the latest release, i.e. *stable*
- use the latest `main`, i.e. *latest*
- use a custom version for each runcard (possibly written in a TOML file [*])
- more?

My proposal would be to use *stable* by default, and add the TOML file in case we want to freeze or advance some platform (e.g. to test new Qibolab features, or maintain a platform that for some reason couldn't be compatible with Qibolab *stable*).

[*]: TOML is suitable for configurations, and it has been added to Python standard library, so we could save the dependency on PyYAML

stavros11:
Fixes #44 by adding a simple CI that checks creation of all platforms. The environment and Python version used is matching the `qibo` module in the cluster, to make sure that all platforms here are compatible.

### DIFF
--- a/tests/test_platforms.py
+++ b/tests/test_platforms.py
@@ -1,0 +1,25 @@
+import os
+import pathlib
+
+import pytest
+from qibolab import Platform, create_platform
+
+PATH = pathlib.Path(__file__).parent.parent
+
+
+def platform_names():
+    """Yield the names of all platforms in the repository."""
+    for filename in os.listdir(PATH):
+        split = filename.split(".")
+        if split[-1] == "py":
+            yield split[0]
+
+
+@pytest.mark.parametrize("name", platform_names())
+def test_create(name):
+    """Test that platform ``name`` can be created."""
+    platform = create_platform(name)
+    qubit = next(iter(platform.qubits))
+    rx_pulse = platform.create_RX_pulse(qubit)
+    mz_pulse = platform.create_MZ_pulse(qubit, start=rx_pulse.duration)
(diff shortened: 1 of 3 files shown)


tests/test_platforms.py
```diff
@@ -0,0 +1,25 @@
+import os
+import pathlib
+
+import pytest
+from qibolab import Platform, create_platform
+
+PATH = pathlib.Path(__file__).parent.parent
```
AleCandido, line 7:
```suggestion
PATH = pathlib.Path(__file__).parents[1]
```

tests/test_platforms.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def platform_names():
+    """Yield the names of all platforms in the repository."""
+    for filename in os.listdir(PATH):
+        split = filename.split(".")
+        if split[-1] == "py":
+            yield split[0]
```
AleCandido, line 15:
```suggestion
    return PATH.glob("*.py")
```

AleCandido (approved):

scarrazza (approved):

PR #55, opened 2023-09-20
Fix resonator 3D -> 2D

MatteoRobbiati:


tii1q_b1.py
```diff
@@ -37,4 +37,4 @@ def create(runcard_path=RUNCARD):
 
     instruments = {controller.name: controller}
     settings = load_settings(runcard)
-    return Platform(NAME, qubits, pairs, instruments, settings, resonator_type="3D")
+    return Platform(NAME, qubits, pairs, instruments, settings, resonator_type="2D")
```
stavros11, line 40: Why is this change required? I think tii1q_b1 is a 3D cavity by fabrication.

MatteoRobbiati: I was getting weird results in a `qubit_spectroscopy` and together with @Edoardo-Pedicillo we noticed the results were fixed setting `resonator_type="2D"`. Maybe I misunderstood  @Edoardo-Pedicillo?

stavros11: We discussed this with @Edoardo-Pedicillo yesterday and the proper fix is not changing the resonator type (even though this may work temporarily). The resonator type affects the `resonator_spectroscopy` where you should get a max for 3D and min for 2D. For the `qubit_spectroscopy` you may get either max or min regardless of the resonator type, depending on the readout frequency you are using. 

This cavity is 3D and I think @andrea-pasquale confirmed that the `resonator_spectroscopy` gives a max. The proper fix for your issue should be done in the fitting of `qubit_spectroscopy` in qibolab (qiboteam/qibocal#528). I will close this PR but feel free to use the branch if you need it (until this is properly fixed).

PR #56, opened 2023-09-26
Add instruments section in runcards

stavros11:
Updates runcards for the new layout implemented in qiboteam/qibolab#585. I would like to wait for qiboteam/qibolab#508 before merging this, because a few more changes may be required in the interface.

I also added a workflow that is running the platform creation test with qibolab main. We can use this to check if the runcards are compatible with main before releasing qibolab. @AleCandido can you help me on how to properly activate the poetry environment in that workflow
https://github.com/qiboteam/qibolab_platforms_qrc/blob/1d0e89d00ce159d5c9fad12bb500721113aa5047/.github/workflows/main.yml#L30

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,16 +17,12 @@ jobs:
         with:
           virtualenvs-create: true
           virtualenvs-in-project: true
-      - name: Disable modern installer
-        run: poetry config installer.modern-installation false
       - name: Install qibolab
         run: |
           git clone https://github.com/qiboteam/qibolab
           cd qibolab
-          poetry install --no-interaction --no-root --with tests --all-extras
-          cd ..
+          poetry install --no-interaction --with tests --all-extras
       - name: Test platforms
         run: |
-          source $VENV
-          export QIBOLAB_PLATFORMS=$(realpath .)
-          pytest tests/
+          cd qibolab
+          QIBOLAB_PLATFORMS=$(realpath ..) poetry run pytest ../tests/
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,32 @@
+name: Tests using qibolab main
+on: push
+jobs:
+  tests:
+    runs-on: ubuntu-22.04
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: Install and configure poetry
+        uses: snok/install-poetry@v1
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+      - name: Disable modern installer
+        run: poetry config installer.modern-installation false
+      - name: Install qibolab
+        run: |
+          git clone https://github.com/qiboteam/qibolab
+          cd qibolab
+          poetry install --no-interaction --no-root --with tests --all-extras
+          cd ..
+      - name: Test platforms
+        run: |
+          source $VENV
+          export QIBOLAB_PLATFORMS=$(realpath .)
+          pytest tests/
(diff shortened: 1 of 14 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: |
+          git clone https://github.com/qiboteam/qibolab
+          cd qibolab
+          poetry install --no-interaction --no-root --with tests --all-extras
+          cd ..
+      - name: Test platforms
+        run: |
+          source $VENV
+          export QIBOLAB_PLATFORMS=$(realpath .)
+          pytest tests/
```
AleCandido, line 32:
```suggestion
          QIBOLAB_PLATFORMS=$(realpath .) poetry run pytest tests/
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+      - name: Disable modern installer
+        run: poetry config installer.modern-installation false
+      - name: Install qibolab
+        run: |
+          git clone https://github.com/qiboteam/qibolab
+          cd qibolab
+          poetry install --no-interaction --no-root --with tests --all-extras
+          cd ..
```
AleCandido, line 27: This is not required at the end of a block (the new shell will re-enter in the default directory).
```suggestion
```

iqm5q_q0.yml
```diff
@@ -10,6 +10,17 @@ qubits: [0]
 
 topology: []
 
+instruments:
+    lo_readout: # For SG0 (Readout)
+        frequency: 5_500_000_000
+        power: null
```
andrea-pasquale, line 16: Is it possible to omit `power` instead of putting `null`?

andrea-pasquale (commented):
Thanks @stavros11, LGTM.
Just a small comment.
As soon as CI passes I will approve

.github/workflows/main.yml
stavros11: This will not work as it is because we need to execute `poetry run` from within the qibolab project directory. Instead I tried 
```sh
cd qibolab
QIBOLAB_PLATFORMS=$(realpath .) poetry run pytest ../tests/
```
 here https://github.com/qiboteam/qibolab_platforms_qrc/actions/runs/6310932947/job/17133888194 but I get that `qibolab` is not installed so maybe the environment is not properly activated.

AleCandido: You're right about the change of directory, the problem with installation is not because of environment activation, but because `qibolab` has never been installed before because of the `--no-root` flag

https://python-poetry.org/docs/cli/

.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+      - name: Disable modern installer
+        run: poetry config installer.modern-installation false
+      - name: Install qibolab
+        run: |
+          git clone https://github.com/qiboteam/qibolab
+          cd qibolab
+          poetry install --no-interaction --no-root --with tests --all-extras
```
AleCandido, line 26:
```suggestion
          poetry install --no-interaction --with tests --all-extras
```

.github/workflows/main.yml
AleCandido: (`--no-root` means: *install all dependencies but not the package itself*)

stavros11: Thanks, I just blindly copied the other workflow without checking what the options were doing. It is working now.

iqm5q_q0.yml
stavros11: Seems possible because the LOs use `None` as default value anyway, thanks.

stavros11:
> Thanks @stavros11, LGTM. Just a small comment. As soon as CI passes I will approve

Thank you for the reviews. The workflow using qibolab main is passing, the other one will require to release qibolab. But I think it makes sense to merge this only after releasing qibolab, in order to maintain the cluster environment.

andrea-pasquale (approved):

AleCandido (approved):

stavros11:
Merging this because I need to make the main workflow available in the other open PRs. We can fix the couplers in a different PR after qiboteam/qibolab#508 is merged.

PR #62, opened 2023-10-09
Selfhosted workflow

stavros11:
I still need to test that this works in practice, but I need to open a PR to do that. The idea is the following

Steps for adding a new platform named `"myplatform"`:
1. `myplatform.py` should be written containing the `create` method,
2. `selfhosted/queues.yml` should be updated mapping the new platform to the corresponding queue (may not have the same name)
3. open a PR with the changes
4. add label `run-on-myplatform` to test on the cluster


selfhosted/actions.yml
```diff
@@ -0,0 +1,13 @@
+platform: MYPLATFORM
+
+qubits: [0]
+
+format: csv
```
andrea-pasquale, line 5: This is no longer necessary. It is not generating any errors but it is better to remove it.
I like a lot the fact that you are using qibocal directly to test platform in CI! In the future we could also use something else other than the single shot classification or potentially more than one protocol.

selfhosted/queues.py
```diff
@@ -0,0 +1,17 @@
+import pathlib
+
+import click
+import yaml
```
AleCandido, line 4: If you save the queues in JSON (the file is extremely simple) and use `argparse` instead of `click`, you can avoid all external dependencies, and just work with the standard library.

selfhosted/calibration.py
```diff
@@ -0,0 +1,46 @@
+import json
+import pathlib
+
+import click
+import yaml
```
AleCandido, line 5: Same of the other one.

But Qibocal already depends on them, so it's less relevant. Nevertheless, still slightly relevant, because you could avoid versions conflict and stuffs like that (but you would have to give up on the standard format of the Qibocal runcard, that has always been YAML).

selfhosted/calibration.py
```diff
@@ -0,0 +1,46 @@
...
+
+    platform = create_platform(name)
+    card["platform"] = name
+    card["qubits"] = qubits = list(platform.qubits)
+
+    acquisition(card, FOLDER, force=True)
+    folder_path = pathlib.Path.cwd() / FOLDER
+    fit(folder_path, update=True)
+
+    results_path = folder_path / "data" / "single shot_0" / "results.json"
```
AleCandido, line 38:
```suggestion
    results_path = folder_path / "data" / "single_shot_0" / "results.json"
```

selfhosted/calibration.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@click.command()
+@click.argument("name", type=str)
+def main(name):
+    card = yaml.safe_load(RUNCARD.read_text())
+
+    platform = create_platform(name)
+    card["platform"] = name
+    card["qubits"] = qubits = list(platform.qubits)
```
AleCandido, line 32: The `qubits` variable seems unused.
```suggestion
    card["qubits"] = list(platform.qubits)
```

selfhosted/calibration.py
stavros11: This needs to be as it is because the qibocal routine `id` is `single shot` (unless we also change that to `single_shot`).

selfhosted/queues.py
stavros11: Actually, if we use JSON maybe we can get read of `queues.py` completely and do it directly in the shell, using `jq` or some [actions trick](https://stackoverflow.com/questions/61919141/read-json-file-in-github-actions). However, `jq` is not available in the cluster, so it is probably not an option, and I am not sure if the actions allow indexing the loaded JSON with a variable (in our case the platform `name`). Also, these solutions are generally harder to implement and read than Python.

If we keep Python, I agree regarding the dependencies.

selfhosted/calibration.py
stavros11: Here, I am not sure if this is as relevant since we are importing qibocal already.

On another related note, we could provide Python API for qibocal, that allows executing routines programmatically, without having to write a YAML. This here is a potential application, but if there aren't many more, maybe it's not very useful to consider (but I think we discussed it in the past).

stavros11:
> I like a lot the fact that you are using qibocal directly to test platform in CI! In the future we could also use something else other than the single shot classification or potentially more than one protocol.

Thank you for the reviews. 

The idea is to have a quick check of the readout fidelities. The test is not entirely correct because it is calibrating and then reporting fidelities, so if the classification parameters in the original runcard are wrong it will not catch it. But I mostly wanted to have a template for future tests. We could test more routines, the main issue is execution time.

selfhosted/calibration.py
AleCandido: > On another related note, we could provide Python API for qibocal, that allows executing routines programmatically, without having to write a YAML. This here is a potential application, but if there aren't many more, maybe it's not very useful to consider (but I think we discussed it in the past).

In a sense, Qibocal already has a Python API, if you check what the CLI is doing is just calling a single function
https://github.com/qiboteam/qibocal/blob/7abc3818fcff5eb78d9df5fe84e5454e10d60989/src/qibocal/cli/_base.py#L53

A nice (and quick) improvement could be to reshuffle the `cli` subpackage into just a `cli` top-level module, and rename the `cli` subpackage to something like `execute`/`run`/..., such that people do not have to access it as `qibocal.cli.autocalibrate` (but rather `qibocal.run.autocalibrate`).
The other option is to even avoid the subpackage and keep everything top-level (but I like grouping :P).
What do you think? @andrea-pasquale

.github/workflows/tests.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: Install qibolab stable
+        if: ${{ matrix.version == 'stable' }}
+        run: |
+          python -m venv testenv
```
AleCandido, line 21: Actually, why do you use a virtual environment?

I would have expected the self-hosted to run in some kind of container, or with some layer of external isolation. Isn't it the case?

.github/workflows/selfhosted.yml
```diff
@@ -0,0 +1,83 @@
...
+          LABELS: ${{ toJSON(github.event.pull_request.labels.*.name) }}
+        run: |
+          echo $LABELS
+          platforms="["
+          combined=""
+          shopt -s lastpipe
+          jq -c '.[]' <<< "$LABELS" | while read label; do
+              platform=(${label//-/ })
+              if [ ${platform[0]} == "\"run" ] && [ ${platform[1]} == "on" ]; then
+                platforms="${platforms}${combined:+,}\"${platform[2]}"
```
AleCandido, line 27: Couldn't you use the `+=` operator also here, as you're doing below?

.github/workflows/tests.yml
stavros11: It seems to fail without the virtual environment

https://github.com/qiboteam/qibolab_platforms_qrc/actions/runs/6533280713/job/17738163307?pr=62

selfhosted/calibration.py
stavros11: > In a sense, Qibocal already has a Python API, if you check what the CLI is doing is just calling a single function https://github.com/qiboteam/qibocal/blob/7abc3818fcff5eb78d9df5fe84e5454e10d60989/src/qibocal/cli/_base.py#L53

Yes, this is what I am also doing here in calibration.py, just calling `acquire` and `fit` from `qibocal.cli` seperately as I don't need the report. But this still requires defining a runcard as a dictionary (or yaml). I was thinking somehow calling directly the routine I want, eg.
```py
from qibocal.routines import single_shot_classification

single_shot_classification(platform, qubits, ...)
```
but maybe this is not very useful. Anyway, this is not very related to this PR and it's not really a priority.

AleCandido: Ok, that is simple enough: my point was just that, since you're serializing a dictionary in your favorite format, you can just use JSON and avoid the further dependency.

No need to call directly single routines, this would skip all the mechanism behind the routine (i.e. storing the acquired data, fit results, ...).

AleCandido (approved):
To me it is minimal and fine. As soon as you're to get back the green tick you can merge.

(I believe the current failures not to be caused by the workflow itself - but it would be nice to merge a working thing nevertheless)

selfhosted/calibration.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def main(name):
+    """Execute single shot classification routine on the given platform."""
+    routine = Operation.single_shot_classification.value
+    params = routine.parameters_type.load(dict(nshots=5000))
+
+    platform = create_platform(name)
+    qubits = platform.qubits
+
```
andrea-pasquale, line 30: I believe that on `qw5q_gold` tests are failing because we are missing the `platform.setup()`, `platform.connect()` and similar...
I forgot to include them in the script that I gave you since I was only testing `dummy`

stavros11: Good point, thanks.

andrea-pasquale (approved):
Thanks @stavros11.
I manage to run on `qw5q_gold` just now and everything seems fine.

PR #64, opened 2023-10-10
Fixing `assignment_fidelity` typo

andrea-pasquale:
Same as qiboteam/qibolab#613.
Now as expected tests with the qibolab latest are passing and qibolab stable are not passing anymore.
If you want both tests to pass we could eventually remove temporarily `assignment_fidelity` from the runcard.

### DIFF
--- a/iqm5q.yml
+++ b/iqm5q.yml
@@ -179,7 +179,7 @@ characterization:
             asymmetry: 0.0
             bare_resonator_frequency_sweetspot: 0.0
             ssf_brf: 0.0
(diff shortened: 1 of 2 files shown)


stavros11 (approved):
Indeed, thanks for the fix. Personally, I don't mind much if the tests are not passing with stable. I just have them there for the time we manage to have a stable user environment in the cluster, but we are not there yet (and currently all users are developers anyway).

AleCandido (approved):

PR #69, opened 2023-10-18
David/iqm5q runcard october2023

DavidSarlle:
Chip characterized after latest warm up

- Single qubit fidelities:

Assg. fidelity q0: 92.4%
Assg. fidelity q1: 98.5%
Assg. fidelity q2: 79.9%
Assg. fidelity q3: 88.7%
Assg. fidelity q4: 96.7% 

T1s:

q0: 11961 ns
q1: 11491 ns
q2: 8343 ns
q3: 10490 ns
q4: 9891 ns


Jacfomg:
Were you able to calibrate the TWPA with the VNA or qibocal ? Because I got values near 0.95 to 0.9 before and I think if more a TWPA thing that anything.

DavidSarlle:
> Were you able to calibrate the TWPA with the VNA or qibocal ? Because I got values near 0.95 to 0.9 before and I think if more a TWPA thing that anything.

Yes I did it with qibocal. The values for freq and powers have changed slightly but as soon as I did it, the assg. fifelities increased from 55 to 72% more or less...

andrea-pasquale:
As soon as we fixed conflicts we should definitely merge

Jacfomg:
All qubits look like this now, as far as I know these kind of streching on a single direction i usually caused during the amplification steps, I would assume the TWPA to is doing something funny.

image

stavros11 (approved):

DavidSarlle:
@stavros11 thanks for the review. But I have detected that the use of the q2 produces a cross talking with all the qubits that decrease the fidelities 20/30% consistenly. The fidelities are pretty good when the q2 is not being used. I am working in a rutine to characterize the cross talk matrix vs. the coupler bias to find the bias values of the couplers where we issolate the qubits o we leave them to interact. Said, that, the runcard is the best one that we have for the moment, but everybody should take into account the cross-talk issue issue.

Jacfomg (approved):
We should merge it so we have some starting point as we will warm up soon.

stavros11:
Thanks @DavidSarlle. I also noticed that the fidelities are not very stable when I tried to use this yesterday. I just approved because it is better than the one on main and also has the right IP for the TWPA. So if you agree you can merge this and you can still keep working on your branch.

andrea-pasquale:
> We should merge it so we have some starting point as we will warm up soon.

+1

andrea-pasquale (approved):

DavidSarlle:
> > We should merge it so we have some starting point as we will warm up soon.
> 
> +1

I agree

PR #73, opened 2023-10-25
bot t1 t2 and RB

Jacfomg:


andrea-pasquale:
Just a comment: if you want to run more than one routine I believe that it is better to switch to the runcard approach.
I can help you with that.

Jacfomg:
> Just a comment: if you want to run more than one routine I believe that it is better to switch to the runcard approach. I can help you with that.

Sure, but the code doesn't look that bad to me.

Jacfomg:
Seems working to me, I will work now on qibocal RB to have 5 gate fidelities and open a PR editing the message and finetuning the RB time.

Jacfomg:
If assigmment fidelity is lower than 0.9 for all do something to skip the RB

stavros11:
My main concern with this is execution time, because 720sec (10min+) is very long for a workflow. Maybe we could keep a shorter workflow with just the classification as the default, to quickly confirm if the runcard is working or no, and have a different label or execution trigger for the extended one.

I see that it is also failing for qblox, I believe because the time sweepers cannot start at 0, should be a minimum of 4ns.

Jacfomg:
> My main concern with this is execution time, because 720sec (10min+) is very long for a workflow. Maybe we could keep a shorter workflow with just the classification as the default, to quickly confirm if the runcard is working or no, and have a different label or execution trigger for the extended one.
> 
> I see that it is also failing for qblox, I believe because the time sweepers cannot start at 0, should be a minimum of 4ns.

I haven't tweeked the RB yet, without it execution was less than 1 minute for T1, T2 and assigment fidelity. Then, for qblox, I was starting at 50ns I don't know why it fails.

Jacfomg:
For the RB seems that the connection to the instruments is really unconsistent and takes longer than usual, from the logs
image
This scan took around 550s while it should have taken 150s.

Jacfomg:
Should I remove the RB for the time being and try to merge it ?

Jacfomg:
Is it ok to keep the RB as commented code for later ?

andrea-pasquale:
> Is it ok to keep the RB as commented code for later ?

Fine by me since you have already found some good initial parameters for a "short" execution

andrea-pasquale:
The only thing is that I am not sure whether this will work on all platforms/instruments

Jacfomg:
It will crash on qblox due to the duration sweepers

andrea-pasquale (approved):

stavros11:
> It will crash on qblox due to the duration sweepers

Are you sure? There is no duration sweeper on T1, T2.
Anyway, we can just test when the instruments are back.

andrea-pasquale:
I think the problem was the sweeper range. At this point we can check as soon as the chips are back and we can give it a try.

Jacfomg:
> Member

Yeah, I think you can still check the qw5q tests failing because of that

.github/workflows/selfhosted.yml
```diff
@@ -77,7 +77,7 @@ jobs:
         source testenv/bin/activate
         export QIBOLAB_PLATFORMS=.
         partition="$(jq -r -n --argjson data "$queues" '$data.'$platform)"
-        srun -p $partition python selfhosted/calibration.py ${{ matrix.platform }}
+        QIBOCAL_LOG_LEVEL=4 QIBO_LOG_LEVEL=3 srun -p $partition python selfhosted/calibration.py ${{ matrix.platform }}
```
stavros11, line 80: Are the log levels still needed? I think we were using them to debug RB.

stavros11 (commented):
I would propose the following changes:
* I understand that the two action runcards added are not related to the selfhosted, but may be useful for the monitoring tool. In that case, we should create a new folder monitoring and put them there, potentially putting different routines per platform in the future.
* I understand the T1, T2 routines may behave differently on different platforms due to sweeper bugs. In that case, I would prefer to enclose the `routine.acquisition` call in a try-except and make sure that the workflow passes in all cases with the bot posting all results from working routines (and possibly saying T2 failed otherwise).

Of course ideally everything should work everywhere, but at least until we manage to do it (which may take some time), it is better to get some feedback, than no feedback. Also implementing this shouldn't be that hard (needs a try-except and some modifications in the `generate_message`), but if you prefer I can take care of it.

Jacfomg:
> I would propose the following changes:
> 
> * I understand that the two action runcards added are not related to the selfhosted, but may be useful for the monitoring tool. In that case, we should create a new folder `monitoring` and put them there, potentially putting different routines per platform in the future.
> * I understand the T1, T2 routines may behave differently on different platforms due to sweeper bugs. In that case, I would prefer to enclose the `routine.acquisition` call in a try-except and make sure that the workflow passes in all cases with the bot posting all results from working routines (and possibly saying T2 failed otherwise).
> 
> Of course ideally everything should work everywhere, but at least until we manage to do it (which may take some time), it is better to get some feedback, than no feedback. Also implementing this shouldn't be that hard (needs a try-except and some modifications in the `generate_message`), but if you prefer I can take care of it.

If you already had a working version with the runcards and you can recover it I would prefer if you could do it.

Jacfomg:
Thanks for implementing this, I would say that having both Readout assignment fidelities and Readout QND it's not needed. And taking into accout we want to test the runcard Readout QND is more suited as it does use the discrimination on the runcard. Also, we can even report the QND, the other value Readout QND reports together with the assigment fidelity. Then for T1/T2 I would prefer to use the values provided in the spreadsheet to do the calculation you mentioned instead of the ones in the current runcard. 

Finally, depending on how precise we want to be I would suggest a smaller but still precise enough scan. That would involve, 25 steps instead of 50 should be ok, 1024 shoots instead of 5000 for the T1/T2 and a scan between 3 to 5 times bigger instead of 10 will also work nicely and reduce the time it take to run the bot.

Jacfomg:
> Run on platform `qw5q_gold_qblox` completed in 44.46sec! :atom:
> 
> Readout assignment fidelities: 0: 0.646 1: 0.501 2: 0.510 3: 0.617 4: 0.498
> 
> Readout QND: 0: 0.510 1: 0.500 2: 0.500 3: 0.503 4: 0.500
> 
> T1: 0: 2.31 us 1: 288.98 us 2: 0.87 us 3: 1.02 us 4: 10.57 us
> 
> T2: 0: 0.02 us 1: 1345.31 us 2: 0.02 us 3: 0.39 us 4: 1342.21 us

Also, is there a way to inspect the reports and see qhat happened in the 1000us T1s ?

andrea-pasquale:
> > Run on platform `qw5q_gold_qblox` completed in 44.46sec! :atom:
> > Readout assignment fidelities: 0: 0.646 1: 0.501 2: 0.510 3: 0.617 4: 0.498
> > Readout QND: 0: 0.510 1: 0.500 2: 0.500 3: 0.503 4: 0.500
> > T1: 0: 2.31 us 1: 288.98 us 2: 0.87 us 3: 1.02 us 4: 10.57 us
> > T2: 0: 0.02 us 1: 1345.31 us 2: 0.02 us 3: 0.39 us 4: 1342.21 us
> 
> Also, is there a way to inspect the reports and see qhat happened in the 1000us T1s ?

I don't know. If the chip is properly calibrated we could also consider to run `t1` and `t2` with probabilities so that you can also have the $\Chi^2$ and the error

Jacfomg:
> Member

That would be quite nice, as I think it's safe to assume the chip should be calibrated

.github/workflows/selfhosted.yml
Jacfomg: Yeah, I already removed them when renaming the files

andrea-pasquale: Ok, I saw the the conversation was unsolved and I decided to reply.. a week later. I will resolve it now.

stavros11 (approved):
Thanks for the feedback. I implemented the suggestions from @Jacfomg for shorter scans.
I also tried using qq-upload within the script so that we can get the reports and while it works from my terminal, in the CI it fails due to a dependency missing (rsync - see here https://github.com/qiboteam/qibolab_platforms_qrc/actions/runs/7183581858/job/19562691384). So I decided to revert to the previous version. I will give a try to fix it and if I manage I will open another PR.
Adding errors is also a good idea but I wouldn't do it yet because it requires single shot which is typically bad. This is just a bot for a quick check of new runcards anyway, for more details we will need to use the monitoring tool or run qq manually.

@Jacfomg @andrea-pasquale if you agree I will merge it as it is. It gives fidelities, T1 and T2 and is relatively short.

stavros11:
@Jacfomg is the following error
```
[2023.12.12 20:11:47.350] CRITICAL Exception Illegal type when calling method set with ([['/dev10055/execution/holdoff', 3.0425600000000004], ['/dev12146/qachannels/0/readout/result/source', 3], ['/dev12146/qachannels/0/readout/multistate/qudits/1/thresholds/0/value', (3.9350260175625893,)], ['/dev12146/qachannels/0/readout/multistate/qudits/2/thresholds/0/value', (0.6334573745826475,)], ['/dev12146/qachannels/0/readout/multistate/qudits/3/thresholds/0/value', (-3.295590970178259,)], ['/dev12146/qachannels/0/readout/multistate/qudits/4/thresholds/0/value', (1.7944859713165011,)], ['/dev12146/qachannels/0/readout/result/enable', 1]],) and {}
```
related to the typo you fixed in https://github.com/qiboteam/qibolab/pull/685 or something different?

GabrielePalazzo (approved):

Jacfomg:
> @Jacfomg is the following error
> 
> ```
> [2023.12.12 20:11:47.350] CRITICAL Exception Illegal type when calling method set with ([['/dev10055/execution/holdoff', 3.0425600000000004], ['/dev12146/qachannels/0/readout/result/source', 3], ['/dev12146/qachannels/0/readout/multistate/qudits/1/thresholds/0/value', (3.9350260175625893,)], ['/dev12146/qachannels/0/readout/multistate/qudits/2/thresholds/0/value', (0.6334573745826475,)], ['/dev12146/qachannels/0/readout/multistate/qudits/3/thresholds/0/value', (-3.295590970178259,)], ['/dev12146/qachannels/0/readout/multistate/qudits/4/thresholds/0/value', (1.7944859713165011,)], ['/dev12146/qachannels/0/readout/result/enable', 1]],) and {}
> ```
> 
> related to the typo you fixed in [qiboteam/qibolab#685](https://github.com/qiboteam/qibolab/pull/685) or something different?

Yes, it was the typo

Jacfomg:
> single shot which is typicall

I wouldn't say singleshot is bad, it takes the same time so we might consider it for the bot as well.

andrea-pasquale:
> Adding errors is also a good idea but I wouldn't do it yet because it requires single shot which is typically bad. This is just a bot for a quick check of new runcards anyway, for more details we will need to use the monitoring tool or run qq manually.
> 
> @Jacfomg @andrea-pasquale if you agree I will merge it as it is. It gives fidelities, T1 and T2 and is relatively short.

Thanks for the updates @stavros11. I think that we can merge.

Jacfomg:
Also, could we add the happy or sad face when results are out of a range we define as well calibrated ?

PR #75, opened 2023-11-01
moved qblox modules settings in runcard

PiergiorgioButtarini:
This modifications have reproduced the basic calibration routine but further testing is needed 
To do:

- [x] Reproduce all routine performance

- [x] General refactoring might be still needed in order to reduce as much as possible the number of new lines in the runcard yaml.
- [x] Dump of the module settings in the runcard 
- [x] Test have to be adjusted

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.


qw5q_gold.yml
```diff
@@ -9,35 +9,126 @@ topology:
...
 instruments:
     twpa_pump:
-        frequency: 6_535_900_000
-        power: 4
+        frequency: 6545500000
+        power: 1.4
+
+    qcm_bb0:
+        o1:
+            channel: L4-5
```
stavros11, line 17: As I wrote in the qibolab PR I have not reviewed the new approach in detail yet, however one suggestion would be to avoid putting the channel here. I understand it may be convenient, however this information already exists in the platform `create` method and we should avoid duplication as much as possible, mostly because it can lead to errors (what if you put a different channel here than in the platform creation?). 
Also, the general idea is that `create` should contain static parameters (channel map, qubit names, instrument names and IPs) which do not change during calibration, while the yaml contains the parameters that may change. Channels do not change (unless you plug/unplug cables) so they should be in `create`.

PiergiorgioButtarini: Thanks for the fast feedback @stavros11, I was planning to do this modification since it is a static and redundant parameter, but for the moment I just reproduced exactly the parameter in the previous approach.
I will delete it and I will also implement the dumping in the runcard.

stavros11:
Superseeded by #91.

Issue #74, opened 2023-10-26
Rename `qw5q_gold_qblox` in `qw5q_gold`


andrea-pasquale:
I think that to make things simple we should have the same names between platforms and partitions in cluster.
If the only reason for maintaining `qw5q_gold` is to store the `qm`configuration we can always recover it looking at the history of the repository.

PR #80, opened 2023-11-08
Alvaro/qw5q gold runcard 20231109 Initial characterisation after cooldown

aorgazf:
Assigment fidelities around 89%

### DIFF
--- a/qw5q_gold.yml
+++ b/qw5q_gold.yml
@@ -77,8 +77,9 @@ characterization:
             Ec: 270000000
             Ej: 11400000000
             g: 107000000
-            assignment_fidelity: 0.8657498362802882
-            readout_fidelity: 0.7314996725605764
+            assignment_fidelity: 0.860347085789129
+            readout_fidelity: 0.720694171578258
+            effective_temperature: 0.0
             peak_voltage: 0
             pi_pulse_amplitude: 0.5037576688262035
             T1: 5662.122123503363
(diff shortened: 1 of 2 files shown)


andrea-pasquale:
Thanks @aorgazf. I just run a `readout_characterization` protocol to assess the quality of the runcard.
Here are the results: http://login.qrccluster.com:9000/qcJPR0cSStu718LJc9qw6A==/
The assignment fidelities with the current runcard are lower than 89% but we can start playing with this runcard for sure.

aorgazf:
Hi @andrea-pasquale, yes I see similar results, I'm not sure why, but it could be due to the fact that the chip is still not cold. This has happened before, in the hours after reaching base temperature.
Additionally, there seems to be a slight problem with the XLD after the maintenance, and I Bluefors engineer and Varun are performing tests that are warming up the fridge a bit temporarily.

stavros11:
Thanks for the runcard. Maybe we should wait for the temperature to stabilize and retry to get better fidelities before we merge.
On another note, we should consider renaming this to qw5q_gold (instead of qw5q_gold_qblox) to fix #74, as I believe it is also easier for the monitoring tool if the platform names are one-to-one with the slurm partitions. There is no qw5q_gold.py file on main anyway.

stavros11:
Qibolab interface has been changed. Superseded by https://github.com/qiboteam/qibolab_platforms_qrc/pull/91.

PR #83, opened 2023-11-13
Use device setup builder api when creating `Zurich` instance

ckutlu:
LabOne Q 2.19 [introduced builder API](https://docs.zhinst.com/labone_q_user_manual/release_notes/#release-2190) for constructing `DeviceSetup` objects at runtime as an alternative to YAML based descriptors. This PR replaces the runtime YAML string assembly present in ZI based platform files.

Since this feature requires LabOne Q 2.19, the dependencies in the related projects should be updated as well.

Currently, this is in draft state. Before merging this,`Zurich` constructor should be modified to consume a `DeviceSetup` instead of a descriptor yaml file.

### DIFF
--- a/iqm5q.py
+++ b/iqm5q.py
@@ -1,6 +1,9 @@
 import itertools
 import pathlib
 
+from laboneq.dsl.device import create_connection
+from laboneq.dsl.device.instruments import HDAWG, PQSC, SHFQC
+from laboneq.simple import DeviceSetup
 from qibolab import Platform
 from qibolab.channels import Channel, ChannelMap
 from qibolab.instruments.oscillator import LocalOscillator
@@ -12,10 +15,6 @@ from qibolab.serialize import (
     load_runcard,
     load_settings,
 )
-
-from laboneq.simple import DeviceSetup
-from laboneq.dsl.device import create_connection
-from laboneq.dsl.device.instruments import HDAWG, SHFQC, PQSC
 
 RUNCARD = pathlib.Path(__file__).parent / "iqm5q.yml"
 
--- a/iqm5q.py
+++ b/iqm5q.py
@@ -38,9 +38,7 @@ def create(runcard_path=RUNCARD):
         HDAWG("device_hdawg", address="DEV8660"),
         HDAWG("device_hdawg2", address="DEV8673"),
         PQSC("device_pqsc", address="DEV10055", reference_clock_source="internal"),
-        SHFQC(
-            "device_shfqc", address="DEV12146"
-        ),
+        SHFQC("device_shfqc", address="DEV12146"),
     )
     device_setup.add_connections(
         "device_shfqc",
--- a/iqm5q.py
+++ b/iqm5q.py
@@ -13,9 +13,15 @@ from qibolab.serialize import (
     load_settings,
 )
 
+from laboneq.simple import DeviceSetup
+from laboneq.dsl.device import create_connection
+from laboneq.dsl.device.instruments import HDAWG, SHFQC, PQSC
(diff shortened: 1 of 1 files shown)


iqm5q.py
```diff
@@ -17,65 +20,79 @@
...
-        {"to": "device_shfqc", "port": "ZSYNCS/2"},
-    ]
+    device_setup = DeviceSetup("EL_ZURO")
+    # Dataserver
+    device_setup.add_dataserver(host="localhost", port=8004)
+    # Instruments
+    device_setup.add_instruments(
+        HDAWG("device_hdawg", address="DEV8660"),
+        HDAWG("device_hdawg2", address="DEV8673"),
+        PQSC("device_pqsc", address="10055"),
```
GabrielePalazzo, line 40:
```suggestion
        PQSC("device_pqsc", address="DEV10055", reference_clock_source="internal"),
```

iqm5q.py
```diff
@@ -17,65 +20,79 @@
...
+        SHFQC(
+            "device_shfqc", address="device_shfqc", reference_clock_source="internal"
```
GabrielePalazzo, line 42:
```suggestion
            "device_shfqc", address="DEV12146", reference_clock_source="internal"
```

iqm5q.py
```diff
@@ -17,65 +20,79 @@
...
-    }
+    device_setup.add_connections(
+        "device_pqsc",
+        create_connection(to_instrument="device_hdawg2", ports="ZSYNCS/1"),
+        create_connection(to_instrument="device_hdawg", ports="ZSYNCS/0"),
+        create_connection(to_instrument="device_shfqc", ports="ZSYNCS/2"),
+    )
 
+    # FIXME: Zurich should be modified to accommodate for device_setup instead of
+    # descriptor
```
GabrielePalazzo, line 91:
```suggestion
```

iqm5q.py
```diff
@@ -17,65 +20,79 @@
...
     controller = Zurich(
         "EL_ZURO",
-        descriptor,
+        # descriptor,
```
GabrielePalazzo, line 94:
```suggestion
```

GabrielePalazzo (changes requested):
Hi @ckutlu, thank you!
Seems good to me. I'll submit a PR on `qibolab` as soon as possible.

iqm5q.py
ckutlu, line 42:
```suggestion
            "device_shfqc", address="DEV12146"
```

ckutlu: I mistakenly included `reference_clock_source` for shfqc, so fixing the device address in a separate suggestion.

GabrielePalazzo (approved):
Thank you very much!
I just tested it and it seems ok to me.

iqm5q.py
```diff
@@ -17,65 +20,74 @@
...
+        create_connection(to_instrument="device_shfqc", ports="ZSYNCS/2"),
+    )
 
     controller = Zurich(
         "EL_ZURO",
-        descriptor,
+        device_setup=device_setup,
         use_emulation=False,
         time_of_flight=75,
         smearing=50,
```
ckutlu, line 93: @GabrielePalazzo I think the following interface would be neater. What do you think? I also left a comment in the parallel PR: https://github.com/qiboteam/qibolab/pull/664

``` py
controller = Zurich.from_device_setup(device_setup, "EL_ZURO", use_emulation=False, time_of_flight=75, smearing=50)
```

Jacfomg (approved):
Looks good, thanks!

PR #86, opened 2023-11-16
Change platform name

Edoardo-Pedicillo:
The name of the platform should be consistent with the file name to use qibocal operational modes.


Jacfomg (approved):
Also, the goal of the iqm5q_q0 is to have a always calibrated q0 for iqm as a first step on the calibration for people to use as fast as possible. As this is not done anymore I would remove it.

stavros11 (approved):
I agree with removing iqm5q_q0.

PR #87, opened 2023-11-20
Refactor qblox channels

PiergiorgioButtarini:
This PR is linked with qibolab PR https://github.com/qiboteam/qibolab/pull/665 and it eliminate the necessity, for Qblox modules, of defning and calling `instantiate_module(modules, cls, name, address, settings)` inside the ` create() ` method, simplifying it and making the Qblox runcards more similar to the others. 

Note: this should be merged immediately after PR #75


qw5q_gold_qblox.py
```diff
@@ -157,85 +34,82 @@ def create(runcard_path=RUNCARD):
...
+
+    instruments = load_instrument_settings(runcard, instruments)
+
+    modules["qcm_bb0"] = qcm_bb0
+    modules["qcm_bb1"] = qcm_bb1
+    modules["qcm_rf0"] = qcm_rf0
+    modules["qcm_rf1"] = qcm_rf1
+    modules["qcm_rf2"] = qcm_rf2
+    modules["qrm_rf_a"] = qrm_rf_a
+    modules["qrm_rf_b"] = qrm_rf_b
```
AleCandido, line 84: Couldn't you generate `modules` just by filtering `instruments`? Are they conceptually different from each other?

PiergiorgioButtarini: Thanks @AleCandido for the tip, I was focused on other thins and I didn't thought about it. Of course you can, I will modify this asap.

AleCandido:
@PiergiorgioButtarini thank you for the PR in general, I was reading it and I found that possible improvement. But the PR is a much bigger improvement by itself :)

qw5q_gold_qblox.py
```diff
@@ -157,85 +34,82 @@ def create(runcard_path=RUNCARD):
...
     #     os.makedirs(folder)
     # for name in modules:
     #     modules[name]._debug_folder = folder
+    qcm_bb0 = ClusterQCM_BB("qcm_bb0", f"{ADDRESS}:2", cluster)
+    qcm_bb1 = ClusterQCM_BB("qcm_bb1", f"{ADDRESS}:4", cluster)
+    qcm_rf0 = ClusterQCM_RF("qcm_rf0", f"{ADDRESS}:6", cluster)
+    qcm_rf1 = ClusterQCM_RF("qcm_rf1", f"{ADDRESS}:8", cluster)
+    qcm_rf2 = ClusterQCM_RF("qcm_rf2", f"{ADDRESS}:10", cluster)
+    qrm_rf_a = ClusterQRM_RF("qrm_rf_a", f"{ADDRESS}:16", cluster)
+    qrm_rf_b = ClusterQRM_RF("qrm_rf_b", f"{ADDRESS}:18", cluster)
```
AleCandido, line 59: Just to tell: in principle this could be also generated, rather than enumerated:
```suggestion
	clusters = {name: ClusterQCM_BB(name, f"{ADDRESS}:" + str(2*(i+1)), cluster) for i, name in enumerate([...])}
```

I believe, only in py>=3.12, you could write:
```suggestion
	clusters = {name: ClusterQCM_BB(name, f"{ADDRESS}:{2*(i+1)}", cluster) for i, name in enumerate([...])}
```

PiergiorgioButtarini: Yes but this is true only if, as in this case, you connect the modules in the even slots of the cluster. Often they are connected pretty randomly, so it must be taken into account the possibility of random or not so even connection

AleCandido: I'm aware that some of these simplifications are working only in specific setups. That's why the people in (or close to) the lab should provide feedback about them.

Moreover, I made an error, since there are two different cluster classes involved (`_RF` and `_BB`).

However, the information here could be slightly compressed anyhow. But most likely it's not convenient in this case.
In case it will be in the future, we could reconsider :)

stavros11 (commented):
Thanks @PiergiorgioButtarini. I believe after merging qiboteam/qibolab#676, we should merge this and close #75 and #80.

PiergiorgioButtarini:
@stavros11 I removed the the offset setting in the runcard but the ports still need to be there since is the way in which the module initialize and know what ports are used.
These three way are equivalent, which one you think I should keep?
``` 
qcm_bb1:
    o1: True
```
``` 
qcm_bb1:
    o1: "connected"
```
``` 
qcm_bb1:
    o1: 
```

stavros11:
Thanks for the update @PiergiorgioButtarini. I was also going to ask in the corresponding qibolab PR why this `o1: True` is needed, as in is there any case where it is `False` (or disconnected)?

Ideally the YAML should contain only parameters that can be calibrated (so mainly numerical values). Whether a port is connected or no is static so it should be in the create method. When you do
```py
channels |= Channel(name="L4-5", port=modules["qcm_bb0"].ports["o1"])
```
you are saying that this port is connected somewhere, we shouldn't need to repeat this again in the yaml. Similarly, if a port is not connected it will not appear at all in the `create`.

In practice, if it is an easy fix to not have these ports in the yaml at all I would go for it, otherwise we can leave it for the next milestone and keep the bool for now.

PiergiorgioButtarini:
> ```python
> channels |= Channel(name="L4-5", port=modules["qcm_bb0"].ports["o1"])
> ```
> 
> you are saying that this port is connected somewhere, we shouldn't need to repeat this again in the yaml. Similarly, if a port is not connected it will not appear at all in the `create`.


 The problem is that in order to be able to call the attribute `modules["qcm_bb0"].ports["o1"]` one should have already defined that `.ports["o1"]` exist and is connected (this happen during the setup starting from the settings read from the yaml). 
For sure there is a way to avoid this but for now I would suggest to merge this, since the main of qibolab and the main of platform_qrc are now decoupled, and I will fix it in the next milestone.

AleCandido:
> The problem is that in order to be able to call the attribute `modules["qcm_bb0"].ports["o1"]` one should have already defined that `.ports["o1"]` exist and is connected (this happen during the setup starting from the settings read from the yaml).

Python has infinite runtime magic (with little effort - sometimes a very unfortunate event), so we can solve the way we prefer.

Actually, the sensible candidates I have for this are the following:
1. the one you wrote
	```py
    modules["qcm_bb0"].ports["o1"]
	```
    and I suggested in https://github.com/qiboteam/qibolab/pull/686#discussion_r1413753397 we could have a `defaultdict(lambda: None)` assigned to `ports`, and using `ports` keys as the list we care about
2. we could avoid the overhead of a `dict` of which we don't need the values (in other languages, like Go, this is the usual way to go to implement sets), but a list/set is not indexed by name, so we need to lift this to a more general object, in which we write the `__getitem__` ourselves, e.g.
	```py
    modules["qcm_bb0", "o1"]
	```
    (if a tuple, we register the element in `.ports`, if not already present)
3. if we don't like to override that `__getitem__`, because we want to keep the original behavior, we could use a different operator, e.g. `__call__`
	```py
    modules("qcm_bb0", "o1")
	```

stavros11:
Or just have a `ports` method in the module and access the ports as
```py
modules["qcm_bb0"].ports("o1")
```
Maybe you still want to have a `_ports` dictionary internally to implement this but we don't need to expose this to the user.

PR #88, opened 2023-11-24
Javi/tii1q_b1 runcard and platform updated 20231124

JavierSerranoGarcia:
Updated runcard after the warm-up. Now the platform includes the twpa local oscilator.


MatteoRobbiati:
Thanks @JavierSerranoGarcia for opening this. I tried to run something on the qubit, but seems it is not calibrated. Is it correct? 

As an example, in this fit the purple line should follow the blue one:
image

Anyway, everything runs properly.

JavierSerranoGarcia:
The performance of the qubit is terrible. I got fidelities next to 0.7 but still, the qubit is useless for computing. 
Now the platform includes the twpa's control so you can run the routines for calibrating the twpa.

stavros11:
Closing this as I don't think this qubit is available anymore.

PR #89, opened 2023-11-30
Updated sweetspot for qubit 0

GabrielePalazzo:
Improves assignment fidelity from 51% to 84% (blobs are still gaussian).


PiergiorgioButtarini (approved):
Thanks, it works fine.

Jacfomg (approved):

PR #90, opened 2023-12-05
Kernels folder and example

Jacfomg:
Add Kernel folder path and file for q0. Should work with https://github.com/qiboteam/qibolab/pull/629.


Jacfomg:
Tests are failing due to qblox @PiergiorgioButtarini any idea why ?

AleCandido:
> Tests are failing due to qblox @PiergiorgioButtarini any idea why ?

Now they are working. After the release, even *stable* will work.

iqm5q.py
```diff
@@ -21,6 +21,7 @@
 TWPA_ADDRESS = "192.168.0.35"
 
 N_QUBITS = 5
+KERNEL_FOLDER = "qibolab_platforms_qrc/iqm5q_kernels/"
```
andrea-pasquale, line 24:
```suggestion
KERNEL_FOLDER =pathlib.Path(__file__).parent/ "iqm5q_kernels"
```
I think in this way it should be better.

iqm5q.py
```diff
@@ -17,6 +17,7 @@
 )
 
 RUNCARD = pathlib.Path(__file__).parent / "iqm5q.yml"
+KERNEL_FOLDER = pathlib.Path(__file__).parent / "iqm5q_kernels/"
```
AleCandido, line 20: Let's just use a single folder for all extra stuffs
```suggestion
FOLDER = pathlib.Path(__file__).parent / "iqm5q"
```

Edoardo-Pedicillo (approved):

stavros11 (approved):
We should merge the qibolab PR first and make sure that the workflow using main passes.

iqm5q.yml
```diff
@@ -228,8 +228,9 @@ characterization:
             state1_voltage: 0
             mean_gnd_states: [-0.0051689363924847615, 0.06828212233945871]
             mean_exc_states: [-0.773457961897901, -0.44969928873718545]
-            threshold: 0.43144500370808087
-            iq_angle: 2.5483919327086055
+            threshold: 0.501546
+            iq_angle: 2.672
+            kernel_path: "kernel_q0"
```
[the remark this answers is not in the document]
Jacfomg: I think it makes more sense in singular, thanks for spotting it.

PR #105, opened 2024-01-12
Use `controller.ports`

stavros11:
In qiboteam/qibolab#739 I removed `Controller.__getitem__` and left only `Controller.ports` to access the ports for consistency, particularly because for some instruments `.ports` has more than one arguments.


AleCandido (approved):

Jacfomg (approved):
Just tried it for iqm, it works

PR #106, opened 2024-01-16
Add support for loading kernels in create() function

Jacfomg:
To work with https://github.com/qiboteam/qibolab/issues/754


iqm5q.py
```diff
@@ -24,7 +25,7 @@
 N_QUBITS = 5
 
 
-def create(runcard_path=RUNCARD):
+def create(runcard_path=RUNCARD, with_kernels: bool = True):
```
AleCandido, line 28: Is there a case in which you want to avoid loading?

iqm5q.py
```diff
@@ -180,7 +181,15 @@ def create(runcard_path=RUNCARD):
 
     # create qubit objects
     runcard = load_runcard(runcard_path)
-    qubits, couplers, pairs = load_qubits(runcard, FOLDER)
+    qubits, couplers, pairs = load_qubits(runcard)
+    if with_kernels and (FOLDER / "kernels.npz").is_file():
+        kernels = Kernels.load(path=FOLDER / "kernels.npz")
+        for q in kernels.data.keys():
```
andrea-pasquale, line 187:
```suggestion
        for q in kernels.data:
```
I think this is not needed.

iqm5q.py
```diff
@@ -180,7 +181,15 @@ def create(runcard_path=RUNCARD):
...
+            # To handle Kernel.save() using strings
+            if q in qubits.keys():
```
andrea-pasquale, line 189:
```suggestion
            if q in qubits:
```
Also this one.

iqm5q.py
```diff
@@ -180,7 +181,15 @@ def create(runcard_path=RUNCARD):
...
+                qubits[q].kernel = kernels[q]
+            else:
+                qubits[int(q)].kernel = kernels[q]
```
andrea-pasquale, line 192: Why here we need to check both cases?

andrea-pasquale: Nevermind I just saw that this for `Kernel.save`.

iqm5q.py
Jacfomg: Maybe you may want to still discriminate with threshold and angle. I am not aware of any case you would want to avoid them, theoretically they should provide better performance, I don't know if they actually do in our lab conditions.

AleCandido: I'd be slightly bolder, and assume that if kernels are available, you should use them (dropping this further argument).
So, the choice of using kernels will depend on the platform, but not on the platform usage.

This will keep this interface simple :)

andrea-pasquale: I agree with @AleCandido. Anyway if they are not available we can replace them it an array of 1s, right @Jacfomg?

Jacfomg: This array of 1s was already being generated inside the zurich driver both for integration and its use with threshold and angle for discrimination. i think there is no need to use the `Kernels` for this and dump the arrays of 1s, right ?

PR #108, opened 2024-01-26
New folder layout

andrea-pasquale:
This PR implements the new folder layout introduce in https://github.com/qiboteam/qibolab/pull/758.
I've also renamed the qblox classes according to https://github.com/qiboteam/qibolab/pull/766 since it got merged in main.
Tests are passing locally running with https://github.com/qiboteam/qibolab/pull/758.


andrea-pasquale:
I forgot to upload the YAML files :(

iqm5q.yml
alecandido, line 1: This has been deleted, while it should have been moved to the corresponding folder

alecandido:
Other than IQM, everything else looks alright. Thanks :)

Jacfomg:
With this I think #106 would not be needed anymore right ?

andrea-pasquale:
> With this I think #106 would not be needed anymore right ?

Yes, I could have also started from that but I didn't see it. Feel free to close it.

Jacfomg:
maybe for the future but this could also look nicer if we put all of them on a qpus folder ?

alecandido:
> maybe for the future but this could also look nicer if we put all of them on a qpus folder ?

I was thinking of renaming all the other folders prepending a `_`. I will do in a subsequent PR.

alecandido (approved):

Jacfomg (approved):
👍

iqm5q/platform.py
```diff
@@ -9,22 +9,22 @@
...
-FOLDER = pathlib.Path(__file__).parent / "iqm5q/"
+FOLDER = pathlib.Path(__file__).parent
 
 TWPA_ADDRESS = "192.168.0.35"
 
 N_QUBITS = 5
 
 
-def create(runcard_path=RUNCARD):
+def create(folder: pathlib.Path = FOLDER):
```
stavros11, line 27: Following https://github.com/qiboteam/qibolab/pull/758#discussion_r1467268557, I am not sure if passing the path as an argument to these functions has any use case. Before it could be used to load the runcard YAML from a different place, that's not very relevant anymore given that the runcard lies in a folder with the `create`. In practice, if someone wants to load a different runcard it is simpler to create a new platform folder (with new create+runcard) altogether.

@alecandido @andrea-pasquale what do you think?

The suggestion is to drop `folder` from argument and just use the constant `FOLDER` directly in the body.

alecandido: Agreed.

I believe the mentioned option has never been used, and if you want to have multiple parameters (e.g. different stages in calibration, or previous calibrations...), you could keep them all in the folder, and just rename the one you want to use.

andrea-pasquale: Fine also with me.

stavros11:
@andrea-pasquale any idea why the tests are still failing with qibolab main, since we merged the other PR? I do not get these failures locally.

andrea-pasquale:
> @andrea-pasquale any idea why the tests are still failing with qibolab main, since we merged the other PR? I do not get these failures locally.

I was going to ask the exact same thing @stavros11 . Also in my case tests are passing locally.
The only difference is that locally I only get 6 instances while here they seems to be 12.
image

stavros11:
Yes, this is what I don't understand. The correct should be 6 since we have 6 platforms in this branch. Where does the CI find the folders named qibolab and dummy...

PR #109, opened 2024-01-26
Prefix service folders with underscore

alecandido:
To separate them from the runcards folders.


alecandido:
@scarrazza I updated the workflows accordingly, but the monitoring tool will have to be updated as well.

scarrazza:
@alecandido no problem, we keep the monitoring tool on hold until this, qibolab and qibocal are operative with the new structure.

alecandido:
> I think I like the `qpus` folder option from [#108 (comment)](https://github.com/qiboteam/qibolab_platforms_qrc/pull/108#issuecomment-1911666060) a bit more, but anything that works is fine.

My motivation to avoid the `qpus` folder was to skip one extra layer, since this is already meant to be the example for how to write runcards. It's just that we need some extra content for our own management.

However, if you have arguments in favor of `qpus`, we could even go for that.

andrea-pasquale:
> > I think I like the `qpus` folder option from [#108 (comment)](https://github.com/qiboteam/qibolab_platforms_qrc/pull/108#issuecomment-1911666060) a bit more, but anything that works is fine.
> 
> My motivation to avoid the `qpus` folder was to skip one extra layer, since this is already meant to be the example for how to write runcards. It's just that we need some extra content for our own management.
> 
> However, if you have arguments in favor of `qpus`, we could even go for that.

I am not sure about folders with `_` but I agree with the fact another layer like `qpus` is probably redundant

stavros11 (approved):
Since most people prefer the current structure I think we can merge this. Just to make sure, the `_` is only for cosmetic reasons, ie. to easily distinguish actual platforms from utilities, right?

alecandido:
> Just to make sure, the `_` is only for cosmetic reasons, ie. to easily distinguish actual platforms from utilities, right?

Exactly, such that the file system and similar interfaces will sort them separately.

andrea-pasquale (approved):

PR #110, opened 2024-01-29
JSON runcards

andrea-pasquale:
This PR modifies all platforms by providing a JSON serialization

- [x] Add `parameters.json` for `qw5q_gold`


alecandido (approved):

alecandido:
Given that this won't work with Qibolab 0.1.5, should we merge this in `main` now?

andrea-pasquale:
Yes, I think that also from qibocal everything should be fine. Perhaps it would be good to add tags also for this repo. This would improve the synchronization with qibolab and qibocal.

stavros11 (approved):
I also prefer merging because, even though not typical and recommended, most people using these repositories work on top of main instead of releases (that's also why the double workflow).

I fully agree regarding the tags, maybe we should create a tag to map current main with qibolab 0.1.5 release before merging this PR. @alecandido can you take care of this (I would do it myself, but I am not sure if I'll do it properly).

alecandido:
> I fully agree regarding the tags, maybe we should create a tag to map current main with qibolab 0.1.5 release before merging this PR. @alecandido can you take care of this (I would do it myself, but I am not sure if I'll do it properly).

I certainly can, and I'm about to, but it just consists in:
```sh
git tag v0.1.5 -m "..."
git push --tags
```

andrea-pasquale:
> @andrea-pasquale @stavros11
> 
> However, I agree with you about tagging for the Qibolab version. But, pay attention that we're facing a multidimensional problem:
> lab config / Qibolab version 	... 	v0.1.5 	v0.1.6 	v0.2.0 	...
> ... 					
> **100** 		x 			
> **101** 		x 	x 		
> **102** 			x 	x 	
> ... 					
> 
> So, we'd need a scheme also to _version_ the lab config, and to link the runcards tags to both of them.
> 
> For the time being, let's keep it simple.
> 
> P.S.: since we might have new runcards for a given (released) version of Qibolab, notice that nothing is immutable in Git, and we can move tags forward, labeling the most recent configuration compatible with a given Qibolab version

Although this is true I believe that `qibolab_platforms_qrc` is mainly driven by `qibolab`. Moreover, for every release of qibolab there is usually a corresponding release of `qibo` and `qibocal`. For now it should be fine to keep the same versioning used in qibolab.

PR #123, opened 2024-02-23
Platform and parameters for the 1q xld1000 platform

JavierSerranoGarcia:
New platform for the tiiqs_xld1000 chip controlled by the rfsoc4x2. Tha characterization is a bad one.


igres26:
Newly calibrated parameters. Some issues during calibration:

- punchout tens to FAIL. It shows an error most of the times that you try to run it. And it also can make the FPGA unresponsive and in need of a reset.
- This same error can appear on some of the other routines but less likely. (THIS NEEDS TO BE FIXED/LOOKED AT)
- T2 sometimes fails (for a different reason) but it can run without problems some times. This is also something that we should check.
- The data quality is good and the times of execution are on par and even better than with the other electronics from this brief experience. 
- Single shot readout is bad. I suspect we might need more amplification on that line, or another look at how the fpga integrate the signal, or some time of flight calibration.

stavros11:
Closing this because this chip is not available anymore. If it is needed in the future we can refer to this PR.

PR #126, opened 2024-03-11
David/iqm5q march 2024

DavidSarlle:
Latest chracterization of iqm5q chip (1q gates).

image

image

image

image

image

When all the qubits are operated at the same time the fidelities decrease a little bit until we chracterized the couplers:

- q0: 80%
- q1: 79%
- q2: 76%
- q3: 82%
- q4: 83%


iqm5q/platform.py
```diff
@@ -152,10 +152,11 @@ def create():
 
     # flux
     for i in range(6, 11):
-        channels[f"L4-{i}"].power_range = 0.8
+        channels[f"L4-{i}"].power_range = 9
```
hay-k, line 155: This is not a valid value for power range for this instrument. This is HDAWG, and for it the power range seems to be in volts (while for the other instruments it is DB). The valid values are `200mV, 400mV, 600mV, 800mV, 1V, 2V, 3V, 4V, 5V`.

For testing my changes in https://github.com/qiboteam/qibolab/pull/848 I used 1V, and things worked perfectly fine.

iqm5q/platform.py
```diff
@@ -152,10 +152,11 @@ def create():
...
+
     # flux couplers
     for i in range(11, 15):
-        channels[f"L4-{i}"].power_range = 0.8
+        channels[f"L4-{i}"].power_range = 9
```
hay-k, line 159: same here

DavidSarlle:
@hay-k thnaks for the review. I set 9 because I coud not find in the docuemntoation what was the maximum value accepted in the PR and it is rounded to the higher allowed value. So, thanks for tehj comment, I will write a comment in this platform section for future characterization as I did with the PR values for the drive for example.

DavidSarlle:
I have checked again this runcard after the latest warm up and cool down and seems to be good. All the fidelities for single qubit are more or less the same (in fact are better for q1 95% and q2 85%). So, from my side this runcard can be merged into main.

DavidSarlle:
Refined all single qubit characterization: 

Single shot classify (all qubits together) decreases a little bit as expected:
http://login.qrccluster.com:9000/LyKcYZnZTnCgADpOkuQK_w==

If the classifier is run qubit by qubit the fidelities are much better as expected:
<html>
<body>

qubit | RO freq | Drive Freq | Sweet Spot | T1 (ns) | T2 (ns) | Assg. Fidelity
-- | -- | -- | -- | -- | -- | --
q0 | 5228650000 | 4103174000 | -0.072 | 19000 | fit fails | 83%
q1 | 4933092000 | 4252836000 | 0.0 | 11100 | fit fails | 94%
q2 | 6106000000 | 4538318000 | 0.0 | 10089 | 972 | 87.6%
q3 | 5803500000 | 4164721000 | 0.0 | 29900 | 160 | 88%
q4 | 5521000000 | 4353668000 | 0.0 | 5800 | 360 | 87%



</body>
</html>

hay-k (approved):

stavros11 (approved):
Since this is tested and working, we could consider merging to have available on main (in light of potential version updates).

PR #138, opened 2024-05-13
initial runcard for iqm5q connected to qblox

aorgazf:


hay-k:
This PR replaces the platform definition. Does this mean that the instrumentation change is intended to be permanent/long term?

DavidSarlle:
@hay-k I am not sure if it is going to be a permanent solution, but the reasons for the change are mainly:

1. Check if there is a problem with the ports/instruments that are controlling the q4 flux needed for the CZs
2. Try to speed up the characterization with better resolution given by Qblox instruments

Also, the change will give us the possibility of compare and check qblox drivers/instruments vs. ZI

I think these are the main reasons for the change, but I suppose that the change will be permanent or long term depending on the results that we observe.

We hope to have results to decided during this week.

aorgazf:
Hi, @DavidSarlle will soon open a new PR with the latest characterisation

PR #139, opened 2024-05-15
[WIP]: tii 2q chip controlled by zi instruments

hay-k:
Experiments involving readout and flux work fine. Have not tested experiments with drive yet - power ranges may need changes.

PR #140, opened 2024-05-15
tii2qs3f with ZI characterisation

biankawolo:
These are parameters and platform for tii2qs3f chip with 2 qubits sharing one resonator. The chip is very fragile and unstable. The characterisation is done up to qubit spectrosopy. My conecern: when you run resonator high power as the first routine and your MZ frequency is slightly different than RO freq you get a resonator peak in different places depending on the difference between RO and MZ values. When RO=MZ freq.  the plot is ok.  If RO = MZ everything is ok.

### DIFF
--- a/tii2qs3f/parameters.json
+++ b/tii2qs3f/parameters.json
@@ -65,9 +65,9 @@
                     "shape": "Rectangular()",
                     "type": "ro",
                     "relative_start": 0
-                }   
+                }
             }
-        },    
+        },
         "two_qubit": { }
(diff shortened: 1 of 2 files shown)


hay-k:
@biankawolo Can you please contribute the changes in parameters to the already existing PR https://github.com/qiboteam/qibolab_platforms_qrc/pull/139 instead of creating a new one.

biankawolo:
oki doki!
________________________________

PR #148, opened 2024-06-27
Recalibrate 11q line D sweetspots after QM hard reboot

stavros11:
We had to hard reboot QM due to an issue with the controllers used in qw5q_platinum. I do not yet understand why, but this changed the sweetspot values that I had calibrated for 11q line D. Here the sweetspots have been recalibrated and fortunately we can still address all qubits with similar assignment fidelities as before:
* Qubit flux: http://login.qrccluster.com:9000/_LCbtU2lQgC5KnZVzK-5fQ==
* Single shot: http://login.qrccluster.com:9000/Vq7uF6HnSgS013NvchHIpQ==
* Rabi: http://login.qrccluster.com:9000/4jXvtyl-QI2QCNejzzc-7A==


alecandido (approved):
Pretty weird, also because the sweetspots really changed a lot (while the classifications parameters are pretty stable...).

Moreover, I don't remember the previous plots, but in the current ones it seems that the flux dependence curve draws a band broader than usual (i.e. if you slice horizontally, the peaks are wider).
Not sure if it's just an impression, and if it is related at all.

andrea-pasquale (commented):
Thanks @stavros11, it looks good to  me.
I was looking at why the flux fit was failing in https://github.com/qiboteam/qibocal/pull/914.
In the runcard, the `drive_frequency` and the `anharmonicity` are missing in the single qubit characterization section. After the fix in https://github.com/qiboteam/qibocal/pull/914 the results are looking better http://login.qrccluster.com:9000/VLJ4dFslQLyPUJl56sIJsQ==/
For the drive frequency and the anharmonicity I could push directly here if you want.

stavros11:
Thanks @andrea-pasquale. Yes, feel free to push. Indeed, when I first created the runcard I put no parameters in the characterization section and then started adding incrementally. Unfortunately, I think we cannot copy the runcard generated by qibocal, due to the issue with string qubit names, so we have to manually copy paste the parameters.

andrea-pasquale:
> Thanks @andrea-pasquale. Yes, feel free to push. Indeed, when I first created the runcard I put no parameters in the characterization section and then started adding incrementally. Unfortunately, I think we cannot copy the runcard generated by qibocal, due to the issue with string qubit names, so we have to manually copy paste the parameters.

Ok, I pushed in https://github.com/qiboteam/qibolab_platforms_qrc/pull/148/commits/ef7824e1d64852ce77022867733011668a57da72. Regarding strings for qubit name, do we really have to use them? I know from @alecandido that you guys are planning to drop them soon from qibolab. We could try to push people to already use integers here to simplify the calibration procedure.
We are ready to merge from my side.

andrea-pasquale (approved):

alecandido:
> Ok, I pushed in [ef7824e](https://github.com/qiboteam/qibolab_platforms_qrc/commit/ef7824e1d64852ce77022867733011668a57da72). Regarding strings for qubit name, do we really have to use them? I know from @alecandido that you guys are planning to drop them soon from qibolab. We could try to push people to already use integers here to simplify the calibration procedure.

I believe this is mostly a purpose from my side. But this will be a non-breaking change: my intention is to preserve them in the runcards, but avoid them inside.
The `QubitId` will be either an index in a list of names hold by the platform (at that point, names could be whatever is JSON serializable, I don't care). Internally, it will be the only name used by Qibolab. But from the outside, you will get translations.

The exact same mechanism could be used in Qibocal. It is not yet clear to me if this should be inherited from Qibolab, or it is better to repeat the translation independently (and I'm leaning towards this second option).

Btw, the implementation of the translation will be trivial:
```py
QubitId = typing.NewType("QubitId", int)

class Platform:
    qubits: list

    def id_to_name(self, id: QubitId) -> Any:
        return self.qubits[id]

    def name_to_id(self, name: Any):
        id = self.qubits.index(name)
        if id < 0:
            raise
        return QubitId(id)
```

stavros11:
Thanks for the reviews. 

Regarding string names, personally I also find them useful in some cases such as this one. Now we are using line A and D, so the natural way to convert to integers is A1, A2, ... A6 -> 0, 1, ..., 5 and D1, ..., D5 -> 6, ..., 10. However, soon we may decide to connect line C and in that case we either map C to 11-15, which is counterintuitive (since C should come before D) or we need to shift all Ds. Either way is confusing for the user. Using strings here also makes it easier to identify which qubits share readout lines and is also in line with the manufacturer labels, so it is easier to write connectivity etc.

This is about the external level only, so basically the platform. Internally in the drivers it is perfectly fine to translate to integers as @alecandido is proposing.

Regarding the issue that we cannot copy the qibocal updated runcard, it may actually be caused by qibolab's serialize module and not qibocal. I will also try to have a look.

PR #154, opened 2024-07-18
Fine tuning D4

andrea-pasquale:
I manage to calibrate `D4` with 99% RB fidelity.
By running a longer RB perhaps we could aim for higher fidelities.
http://login.qrccluster.com:9000/KIEStjkSTwmU-6qNF0nB0g==/


alecandido (approved):
Nice!

stavros11 (approved):

andrea-pasquale:
Perhaps merging now we don't have the same fidelity that I posted before but for sure it is an improvement compared to what we had before. I'm merging now.

stavros11:
> Perhaps merging now we don't have the same fidelity that I posted before but for sure it is an improvement compared to what we had before. I'm merging now.

I implemented a qibocal routine for RB using QUA code that samples the random sequences on the device, which is usually faster and we can do more depths/sequences. Fidelities still look pretty good:
* D1: http://login.qrccluster.com:9000/AZad9bcZQCymhEl58j5WNA==
* D2: http://login.qrccluster.com:9000/s5jxwNo9TcCWN4FJxsLkug==
* D3: http://login.qrccluster.com:9000/W8GG4k_ZTxymoKNpqdNZ0w==
* D4: http://login.qrccluster.com:9000/YohSRfCVTfuMZ0Aseyinzg==

D4 also looks more fine tuned, maybe one more 9 compared to the others.

andrea-pasquale:
Thanks @stavros11, looks very good.
You think we can say that we have three 9s for `D4`? If it is the last value in the table probably yes.

stavros11:
> Thanks @stavros11, looks very good. You think we can say that we have three 9s for `D4`? If it is the last value in the table probably yes.

The three last values in the table are the ones computed here: https://github.com/qua-platform/qua-libs/blob/d530c555ffe3613c6e8715f8eb19fa68272ae734/Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/16a_randomized_benchmarking.py#L314-L316. I am not sure what is typically reported as fidelity in the literature and how 9s are extracted. I guess the last one, but in the end it's the same number from the fit divided by different constants.

PR #156, opened 2024-07-24
Reduce readout power of D2, D3

stavros11:
Reduced a bit the readout pulse amplitude to get better blob shapes (less tails).
* Before: http://login.qrccluster.com:9000/TKkpKJzOSJCZrC_Yvtph0A==
* After: http://login.qrccluster.com:9000/sn2_g81ETz-NGvg41bXkcA==

### DIFF
--- a/qw11q/parameters.json
+++ b/qw11q/parameters.json
@@ -643,7 +643,7 @@
                 },
                 "MZ": {
                     "duration": 2000,
-                    "amplitude": 0.002,
+                    "amplitude": 0.0012,
                     "shape": "Rectangular()",
                     "frequency": 7379568000,
                     "relative_start": 0,
@@ -672,7 +672,7 @@
                 },
                 "MZ": {
                     "duration": 2000,
-                    "amplitude": 0.0011,
+                    "amplitude": 0.0009,
                     "shape": "Rectangular()",
                     "frequency": 7490960000,
                     "relative_start": 0,


andrea-pasquale (approved):
Thanks @stavros11, I believe that you mean `D1` and `D2` in the title, correct?
By running single shot the fidelity you get is with the angle and threshold computed by the fit. At least for `D2` it seems that the difference is quite a lot.
Shall we update also the angle and threshold in `parameters.json`?

stavros11:
> Thanks @stavros11, I believe that you mean `D1` and `D2` in the title, correct? 

Thanks @andrea-pasquale. I think the title is correct, if you see "files changed" the changes are only on `D2` and `D3`, not `D1`.

> By running single shot the fidelity you get is with the angle and threshold computed by the fit. At least for `D2` it seems that the difference is quite a lot. Shall we update also the angle and threshold in `parameters.json`?

Indeed, you are right. I updated the threshold and angle in 52504f40b2ef8a83a465b673c119e0e986161bc2.

stavros11:
I don't know why, but all the sweetspots of line D have shifted significantly, therefore the runcard on main is not correct anymore. I am not sure if this is a temporary shift and will go back to the previous values, but it seems to be like that for several hours so I have recalibrated in 1e9aa9ff406f5884403105e2e1382ef4ae2dc102 (http://login.qrccluster.com:9000/K3H3lUrmRvS2AK3kl7dMuQ==). All other parameters should remain the same.

andrea-pasquale:
@alecandido given that the platform in main is outdated I would merge this one asap.

alecandido:
> @alecandido given that the platform in main is outdated I would merge this one asap.

Sorry, I didn't notice I was the reviewer.

However, in general I'd say it is urgent to merge changes to the `.py`, since without that everything is broken.
Recalibration is never urgent, since everyone is calibrating (unless we're at the stage of giving chips for production... but it didn't seem the case yet...)

alecandido (approved):

stavros11:
Thanks. 

> However, in general I'd say it is urgent to merge changes to the `.py`, since without that everything is broken. Recalibration is never urgent, since everyone is calibrating (unless we're at the stage of giving chips for production... but it didn't seem the case yet...)

That's true, however I would still urge people calibrating to open PRs if they manage to obtain something new or some improvement, even if the changes are small. This way others are aware and can use it, so that we avoid repeating work already done by others as much as possible.

alecandido:
> That's true, however I would still urge people calibrating to open PRs if they manage to obtain something new or some improvement, even if the changes are small. This way others are aware and can use it, so that we avoid repeating work already done by others as much as possible.

Oh yes, definitely.

I have nothing against opening PRs. To me, you should open it as soon as you do anything (if unfinished, just make it draft).
Instead, I was arguing that it is not that urgent to merge them, unless they reflect updates to the lab status, for which the information availability is limited.

The moment QRC (whatever group) will guarantee chips' availability for production, even recalibrating will be a priority. As middleware, the most relevant thing we can do is to make it simple and possibly automated (at that point, even the role of this repo and calibration PRs might be challenged, and we may have to move towards the dedicated non-Git-based database, at least for parameters' values).

@stavros11 @andrea-pasquale

PR #157, opened 2024-07-25
Javi/spinq10q 12 zcu111

JavierSerranoGarcia:
Platform and runcard for Spinq10q chip qubits 2, 3, and 4 controlled with ZCU111. This PR works with qibosoq, qibolab, and qibocal main.


spinq10q.yml
```diff
@@ -0,0 +1,274 @@
+nqubits: 10
```
alecandido, line 1: This information is definitely redundant, since you're listing them anyhow in the `qubits:` field. But maybe it should be 5.

Should we drop this key in 0.2? @stavros11

spinq10q.yml
alecandido, line 1: The current Qibolab version switched to JSON, instead of YAML. Maybe it would be worth to convert this as well (unless you're using quite an old version).

In case, you can do with Python with:
```py
import json
from pathlib import Path

import yaml

PATH = Path("spinq10q.yml")
NEWPATH = Path("spinq10q.json")

NEWPATH.write_text(json.dumps(yaml.safe_load(PATH.read_text())))
```

alecandido (commented):
Thanks for this @JavierSerranoGarcia!

Could you specify somewhere (even in the files) which version of Qibolab and Qibocal they should be used with? If not exactly a version, even the commits are fine

alecandido:
> This PR works with qibosoq, qibolab, and qibocal main.

Ah, sorry, I haven't read it. But I don't expect to work with current `main`. At least not after:
https://github.com/qiboteam/qibolab/commit/83bdae5c601aa5f939ff77168a14bdcc37705661

spinq10q.yml
stavros11: > Should we drop this key in 0.2? @stavros11

Yes, we could drop. However looking at https://github.com/qiboteam/qibolab/blob/c566eace2ba4b6bc165c3411fb8635b31f20a7f9/src/qibolab/platform/platform.py#L126 I believe this already never loaded in the platform, so it would work even if we remove it (I did not test). If that’s the case, it’s just a matter of dropping from the runcards.

stavros11:
> > This PR works with qibosoq, qibolab, and qibocal main.
> 
> Ah, sorry, I haven't read it. But I don't expect to work with current `main`. At least not after: [qiboteam/qibolab@83bdae5](https://github.com/qiboteam/qibolab/commit/83bdae5c601aa5f939ff77168a14bdcc37705661)

I have the impression that the yaml is redundant. There’s also another platform for three qubits committed here that is using json. This may be the one we need and should be compatible with latest main.

JavierSerranoGarcia:
Hi guys, I just want to put in main the platform qibolab_platforms_qrc/spinq10q_12_zcu111. There is no yaml or weird things, just a json file for parameters and python code for the platform, and everything working with the last versions of qibolab and qibocal.

alecandido:
You're both right: I was confused because I started from the first file in the "Files Changed" tab.
This PR is actually adding 4 files, but only those within the folders most likely are relevant, while the other two are related to a supposed Qblox platform (not even sure if it's connected), that is indeed using the YAML.

Should I remove those files from the PR?

spinq10q_12_zcu111/platform.py
```diff
@@ -0,0 +1,94 @@
...
+from qibolab.serialize import (
+    load_instrument_settings,
+    load_qubits,
+    load_runcard,
+    load_settings,
+)
+
+# from qibolab.kernels import Kernels
+
+NAME = "spinq10q_12__zcu111"
```
alecandido, line 16:
```suggestion
NAME = "spinq10q_12_zcu111"
```

Just to match the name of the folder (there was an underscore in excess)

alecandido:
I tried to run, but I'm getting the following error from QICK:
```py
  File "/home/xilinx/qibosoq/src/qibosoq/server.py", line 140, in handle
    results = execute_program(data, self.server.qick_soc)
  File "/home/xilinx/qibosoq/src/qibosoq/server.py", line 103, in execute_program
    toti, totq = program.perform_experiment(
  File "/home/xilinx/qibosoq/src/qibosoq/programs/base.py", line 246, in perform_experiment
    res = self.acquire(  # pylint: disable=E1123,E1120
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/averager_program.py", line 106, in acquire
    avg_d = super().acquire(soc, soft_avgs=self.soft_avgs, load_pulses=load_pulses, start_src=start_src, threshold=threshold, angle=angle, progress=progress)
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/qick_asm.py", line 1145, in acquire
    self.config_all(soc, load_pulses=load_pulses)
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/asm_v1.py", line 672, in config_all
    super().config_all(soc, load_pulses)
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/qick_asm.py", line 668, in config_all
    self.config_readouts(soc)
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/qick_asm.py", line 748, in config_readouts
    soc.configure_readout(ch, output=cfg['sel'], frequency=cfg['freq'], gen_ch=cfg['gen_ch'])
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/qick.py", line 698, in configure_readout
    buf.set_freq(frequency, gen_ch=gen_ch)
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/drivers/readout.py", line 553, in set_freq
    self.readout.set_freq(f, self.readoutport, gen_ch=gen_ch)
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/drivers/readout.py", line 297, in set_freq
    self.set_freq_int(freq_int, in_ch, out_ch)
  File "/usr/local/share/pynq-venv/lib/python3.10/site-packages/qick/drivers/readout.py", line 308, in set_freq_int
    raise RuntimeError("frequency collision: tried to set PFB output %d to %f MHz and output %d to %f MHz, but both map to the PFB channel that is optimal for [%f, %f] (all freqs expressed in first Nyquist zo
ne)"%(out_ch, newfreq, self.out_chs[in_ch], oldfreq, lofreq, hifreq))
RuntimeError: frequency collision: tried to set PFB output 2 to 118.631001 MHz and output 1 to 81.351999 MHz, but both map to the PFB channel that is optimal for [-128.000000, 128.000000] (all freqs expressed
 in first Nyquist zone)
```

spinq10q_12_zcu111/platform.py
stavros11: I would even remove
```suggestion
```
I think this variable is not used anywhere and is even duplicating the `FOLDER` defined in line 21.

stavros11:
> Should I remove those files from the PR?

Yes, I think we should remove.

alecandido:
> Yes, I think we should remove.

Ok, I'm waiting for @JavierSerranoGarcia approval, and I'll remove them.

In the meanwhile, I rebased the branch, since it was still containing old `tii_*` platforms, that are not any longer in `main`.

@JavierSerranoGarcia in case you have troubles pulling, just use instead:
```sh
git pull --rebase
```

JavierSerranoGarcia:
Hi, the only files that should be merged are parameters and platform in the folder spinq10q_12_zcu111. The rest of the file can be removed ...
Regarding the problem of the frequency collisions happens that the resonator are very close in this chip and qick use a polyphase filter with very wide passband and often collisions happens. The trick is using a LO frequency 7.43 GHz for qubits 3 and 4, and LO frequency 7.39 GHz with qubits 2 and 3.

By the way, it would be very useful if we could write these tricks in the parameters.json file.

alecandido:
> The trick is using a LO frequency 7.43 GHz for qubits 3 and 4, and LO frequency 7.39 GHz with qubits 2 and 3.

If I understand correctly, it seems you're talking the two-qubit gates (since 2-3 and 3-4 are the pairs connected in the runcard's topology).
If you'd like to change the LO frequency during execution, we can consider what's the best we could do, but changing the external LO frequency is a considerably slower operation, that can not happen in real-time. Instead, if it is about the pulses' frequency (as it seems, since the frequencies you're quoting are close enough to be in range for the IF), that's of course already available in `parameters.json`. You just need to change the frequency of the natives

However, I got the problem above just running single-shot classification for all qubits at the same time (so no two-qubit gates). If you can slightly adjust the frequencies of the natives to fix the problem above it would be optimal. Otherwise, maybe I should ask you more details about the issue, because I may have misunderstood.

stavros11:
> If I understand correctly, it seems you're talking the two-qubit gates (since 2-3 and 3-4 are the pairs connected in the runcard's topology). 

I don't think the issue is related to two-qubit gates. If I understand it correctly, it is about the LO used for readout, so basically means that we cannot measure all three qubits in the same execution, regardless of what drive gates are used before.

> If you'd like to change the LO frequency during execution, we can consider what's the best we could do, but changing the external LO frequency is a considerably slower operation, that can not happen in real-time. 

Indeed, changing the LO frequency is not possible in real-time. The particular LO used here is external (R&S) so it is certainly not possible. I think that even for other instruments with internal LO, changing its frequency requires some time for the signal to stabilize, so it is not recommended for real-time. Therefore, I don't think there is much we can do from the qibolab side to solve this problem. We already provide interfaces to update LO frequencies in the script but these are not real-time.

> Instead, if it is about the pulses' frequency (as it seems, since the frequencies you're quoting are close enough to be in range for the IF), that's of course already available in `parameters.json`. You just need to change the frequency of the natives

I also don't fully understand what is causing this frequency collision. Usually we need to change LO frequency if the frequencies we want to send are beyond the IF bandwidth, but this doesn't seem to be the case here. Note that the frequency of the natives is mainly dictated by the chip so we cannot change it much if we want to measure properly. We can probably get away with a few Hz-kHZ but not MHz, so I am not sure if we can fix the problem this way. Another solution could be to plug two LOs to the same line, but this will probably create other side effects.

> However, I got the problem above just running single-shot classification for all qubits at the same time (so no two-qubit gates). If you can slightly adjust the frequencies of the natives to fix the problem above it would be optimal.

For now I tried running the single shot on just qubits 3, 4. This does not give the frequency issue, but it gives another one:
```py
qibosoq.client.QibosoqError: Traceback (most recent call last):
  File "/home/xilinx/qibosoq/src/qibosoq/server.py", line 140, in handle
    results = execute_program(data, self.server.qick_soc)
  File "/home/xilinx/qibosoq/src/qibosoq/server.py", line 103, in execute_program
    toti, totq = program.perform_experiment(
  File "/home/xilinx/qibosoq/src/qibosoq/programs/base.py", line 257, in perform_experiment
    return self.collect_shots()[-2:]
  File "/home/xilinx/qibosoq/src/qibosoq/programs/base.py", line 294, in collect_shots
    tot.append(stacked[:, idx].reshape(shape).tolist())
ValueError: cannot reshape array of size 10000 into shape (2,1,10000)
```
I tried with both qibosoq relase and main.

alecandido:
The last issue I believe is related to https://github.com/qiboteam/qibocal/issues/737.

@rodolfocarobene updated Qibosoq after that, and now we're always reshaping. Though I'm not sure why the initial `2`, since it seems just the number of qubits.
I could investigate myself, but maybe @rodolfocarobene already has a clearer picture of what's going on.

JavierSerranoGarcia:
The problem with the frequency collisions is due to the polyphase filter. The bandpass width of one filter in the bank is around 160 MHz and we can only read one qubit per filter. However the resonator  frequencies in our chips are every 50 MHz (approx), too close for these filters. Choosing  the frequency of the LO properly we can get two qubits in different filters and read them simultaneously but not 3. 

The solution is redesigning the polyphase filter and we are working on it. But in the meantime just let's merge this platform.

The problem with the reshaping is something I already commented to Rodolfo.

alecandido:
> The problem with the frequency collisions is due to the polyphase filter. The bandpass width of one filter in the bank is around 160 MHz and we can only read one qubit per filter. However the resonator frequencies in our chips are every 50 MHz (approx), too close for these filters. Choosing the frequency of the LO properly we can get two qubits in different filters and read them simultaneously but not 3.
> 
> The solution is redesigning the polyphase filter and we are working on it. But in the meantime just let's merge this platform.

Ok, so I get your problem is demultiplexing the signal you're receiving on an individual physical channel.
Since you can only split signals in bands of a certain minimum bandwith, you'd like to have a specific LO frequency, such that different qubits are falling in different minimum bands (though it is impossible to disentangle in this way 3 qubits, because you can tune one threshold, but the second one would be too far in any case).

This of course you can do ahead of time: having measured the qubit frequencies, you can compute your desired LO frequency, and set that to your external LO. And it's definitely in your runcard:
https://github.com/qiboteam/qibolab_platforms_qrc/blob/4eea3fbe7f65edeabe9dfaf8cec4c3a4248499e7/spinq10q_12_zcu111/parameters.json#L23-L25

In principle, you can even change dynamically during an experiment, but it makes little sense for Qibolab to switch LO frequency according to the pulse is being given.
What you could do, instead, is to create two different platforms: one for qubit 2 and 3, with LO 7.39 GHz, and one for qubit 3 and 4, with LO 7.43 GHz.
They will both work, and it's not very useful anyhow to have a single platform with all three qubits, since (as you're saying) you can never use all the three of them at the same time.

JavierSerranoGarcia:
Thank you @alecandido, it is a good solution.

alecandido:
@JavierSerranoGarcia this PR is most likely outdated now. Can we close it?

JavierSerranoGarcia:
Hi @alecandido, It is certainly outdated. The ZCU111 is connected to a different chip (but the same wires, so the platform should be OK). Would you like me to create a new, updated platform so you can test code on it?

JavierSerranoGarcia:
THe platform on branch Javi/spinq10q 12 zcu111 is no longer used

alecandido:
> Hi @alecandido, It is certainly outdated. The ZCU111 is connected to a different chip (but the same wires, so the platform should be OK).

Hi @JavierSerranoGarcia, thanks for the update!

> Would you like me to create a new, updated platform so you can test code on it?

For the time being, it is not urgent: we will try to update the driver to the new version of Qibolab. At the beginning, we won't need anything, just to make sure it's barely consistent.

Then, a board in loopback should be sufficient for another while (flashed with QICK tproc v1, non-modified firmware - either the basic or multiplexed firmware should be equally good). We can discuss the optimal connections when we'll get there.

We will only need a qubit connection for the final part (to make sure all routines will be correctly supported). However, if you already have a connected board (with the aforementioned firmware) and that's simpler to you, we can work with that from the beginning.

For as long as it respects the requirements (i.e. v1 non-modified), we should not be sensitive to which specific board, since Qibosoq is supposed to work with all of them (111, 216, 4x2).

JavierSerranoGarcia:
@alecandido, @aorgazf ,  we are maintaining a QPU based on our chips and our control electronics. So, we change the chip and firmware from time to time with better versions and I am raising a PR  for every new version of the platform just in case someone wants to use it.

We are starting to test the tProc V2 and think it will be the standard firmware shortly. So if you are developing any software it should be for this version and, in addition, we are adapting the firmware to fix some issues we are finding in the standard version. Take it into account when you develop something.

alecandido:
> We are starting to test the tProc V2 and think it will be the standard firmware shortly. So if you are developing any software it should be for this version and, in addition, we are adapting the firmware to fix some issues we are finding in the standard version. Take it into account when you develop something.

Hi @JavierSerranoGarcia, I'm aware of it, and perfectly agree with you.

However, I'm just proceeding methodically in steps: as you may know, we have a new major version of Qibolab around (0.2), and the current upgrade is to support this new version.
In order to make minimal changes, we're going to first upgrade the driver, without touching Qibosoq at all.

Of course, after that, the next steps will involve a closer integration with the tProc v2 (on top of a generally closer integration with QICK, than what is currently provided by Qibosoq, as agreed with @aorgazf as well).

> in addition, we are adapting the firmware to fix some issues we are finding in the standard version. Take it into account when you develop something.

That I'm aware of, especially from the meetings with QICK team. There is no single person fully dedicated to that. But as soon as we can get there, we will keep supporting the driver for all the firmware(s) in current use.

PR #160, opened 2024-08-01
Line D for `qw11q` -- Calibration update 

igres26:
Calibration update for Line D in `qw11q`

**NOTE:** High ZZ crosstalk between connected qubits. Ramsey and T2 NEED to be executed separately. It also seems to have an effect on 2-qubit gate calibration. 

It includes slightly changed pre-distortion filters than the ones we have usually been employing.

Some reports for single qubit:

T1: http://login.qrccluster.com:9000/IPVrZtGkTdCX-65eigiovw==

T2: http://login.qrccluster.com:9000/PqzqhkBDQlyuPQWhMNt-oA==

Single shot: http://login.qrccluster.com:9000/4y4uwLpKQtCRAeEiuL30Iw==

RB (using the qua code from https://github.com/qiboteam/qibocal/pull/959): http://login.qrccluster.com:9000/6AVdXbeMSPyswOglW3Q6Fg==/

Some reports for two qubit:

Chevron (to be fixed): http://login.qrccluster.com:9000/KxJTM7wlSeezWEpA6GfbZg==

Tune Landscape D1-D2: http://login.qrccluster.com:9000/xTy6j6_oTPqd6aK-EJ1KLQ==
^Here you can see how the relative angle by the two qubit gate is not the same for both qubits.

### DIFF
--- a/qw11q/parameters.json
+++ b/qw11q/parameters.json
@@ -2,7 +2,7 @@
     "nqubits": 16,
     "settings": {
         "nshots": 2048,
-        "relaxation_time": 100000
+        "relaxation_time": 200000
     },
     "qubits": [
         "A1",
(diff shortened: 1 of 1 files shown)


stavros11:
@andrea-pasquale I would guess that this is also pretty outdated by now?

andrea-pasquale:
Indeed we can close it @stavros11, thanks!

PR #164, opened 2024-08-12
Sweetspot recalibration

andrea-pasquale:
All sweetspots of qw11q shifted significantly after the last calibration:
http://login.qrccluster.com:9000/-knA-9B0RnSIGB2L7-tp2A==

I based this on main instead of #160 since all two qubit gates will also need recalibration.


alecandido:
I'm tagging @aorgazf and @biankawolo, since this issue has been mentioned during last Friday meeting as well.

alecandido (approved):

stavros11 (approved):
Thanks @andrea-pasquale for the recalibration. I believe we should merge this.

> I'm tagging @aorgazf and @biankawolo, since this issue has been mentioned during last Friday meeting as well.

Yes, I noticed that the sweetspots changed last week, sometime between Wednesday and Thursday. Since this has happened before, in the meeting there was the question of whether the new sweetspot values coincide with some previous ones, however comparing this with #156, it does not seem to be the case. It also seems that we can no longer address qubit D5, but this qubit was not working very well even before.

andrea-pasquale:
> Yes, I noticed that the sweetspots changed last week, sometime between Wednesday and Thursday. Since this has happened before, in the meeting there was the question of whether the new sweetspot values coincide with some previous ones, however comparing this with #156, it does not seem to be the case. It also seems that we can no longer address qubit D5, but this qubit was not working very well even before.

Regarding D5, yes, we can in theory still address it. However, putting into its sweetspot generates crosstalk with the other qubits. This is why I decided to put far away from its sweetspot which is around 0.4 - 0.5.

PR #167, opened 2024-08-20
qw5q_platinum calibration

DavidSarlle:
This runcard works with qibocal and qibolab main

q0, q2, q3, q4 single qubit gates calibrated. 
q1 not working

classify

q0 (95.1%): http://login.qrccluster.com:9000/mBBUumF3TUKeWdls4ueGDA==
q1: Qubit not working. 
q2 (95.4%): http://login.qrccluster.com:9000/jnNCfLMKTvuo-tuDHq3wVg== 
q3 (93.8%): http://login.qrccluster.com:9000/eXAUMIbwTkWlR9E9mgyWng==
q4 (93.7%): http://login.qrccluster.com:9000/uO9k8JQCSOCDtQu4Q3FjvQ== 

[WIP]: Running the following routines to fine tuning the qubits:

TWPA freq and pow
Drag pulse

PR #168, opened 2024-08-21
Removing kernels from `qw5q_platinum` platform

andrea-pasquale:
See https://github.com/qiboteam/qibolab_platforms_qrc/pull/167#issuecomment-2301498575
It would be good if someone can reproduce the issue before merging.


qw5q_platinum/platform.py
```diff
@@ -44,8 +44,7 @@ def create():
 
     # create qubit objects
     runcard = load_runcard(FOLDER)
-    kernels = Kernels.load(FOLDER)
```
hay-k, line 47: Instead of just removing the code that loads kernels, I would suggest to load the kernels conditionally based on the existence of the file, or replace the kernel file with an empty kernel file.

andrea-pasquale: Makes sense, now kernels are loaded only if the file exists: https://github.com/qiboteam/qibolab_platforms_qrc/pull/168/commits/413a54f02117f0c264e5d1ad44a19ecd31f5af50

hay-k (approved):

PR #171, opened 2024-08-26
Sweetspot recalibration for qw11q

andrea-pasquale:
I manage to recalibrate only qubits [`D1`](http://login.qrccluster.com:9000/ZZTsob4YQ4utJIycz0q-vg==/), [`D2`](http://login.qrccluster.com:9000/hwpGejQ5RYOJSGA4T9Avug==/) and [`D3`](http://login.qrccluster.com:9000/IhhxbKARRm-sAnUlvJy5MA==/).
Qubit `D4` seems to be out of reach with the latest shifts.

### DIFF
--- a/qw11q/parameters.json
+++ b/qw11q/parameters.json
@@ -791,7 +791,7 @@
                     {
                         "duration": 45,
                         "amplitude": 0.1,
-                        "shape": "Google(0.55, 0)",
+                        "shape": "GaussianSquare(5, 0.9)",
                         "qubit": "D2",
                         "relative_start": 0,
                         "type": "qf"
--- a/qw11q/parameters.json
+++ b/qw11q/parameters.json
@@ -152,6 +152,52 @@
                 "gain": 10
             }
         },
+        "con9": {
+            "o3": {
+                "filter": {
+                    "feedforward": [
+                        1.0684635881381783,
+                        -1.0163217174522334
+                    ],
+                    "feedback": [
+                        0.947858129314055
+                    ]
+                }
+            },
+            "o4": {
+                "filter": {
+                    "feedforward": [
+                        1.0668,
+                        -1.0162
+                    ],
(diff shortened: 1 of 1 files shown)


qw11q/parameters.json
```diff
@@ -743,21 +789,21 @@
             "D1-D2": {
                 "CZ": [
                     {
-                        "duration": 70,
-                        "amplitude": 0.2,
-                        "shape": "GaussianSquare(5, 0.9)",
+                        "duration": 45,
+                        "amplitude": 0.1,
+                        "shape": "Google(0.55, 0)",
```
alecandido, line 794: What is this? xD

alecandido: I may have an idea about that, but it seems to be unsupported by both Qibolab 0.1 and 0.2

andrea-pasquale: Oh sorry I accidentally pushed some of my black magic that I am using to calibrate two qubit gates (without success). I'll remove them.

alecandido (approved):

stavros11 (approved):
Thanks @andrea-pasquale.

> Qubit `D4` seems to be out of reach with the latest shifts.

I only checked the single shot routine, but if this is really an issue we could consider switching to the new QM devices which have a wider range in voltages. However, this would require updating the driver to support them, so I am trying to postpone it as much as possible so that I don't have to do it twice (for qibolab 0.1 and 0.2).

andrea-pasquale:
> I only checked the single shot routine, but if this is really an issue we could consider switching to the new QM devices which have a wider range in voltages. However, this would require updating the driver to support them, so I am trying to postpone it as much as possible so that I don't have to do it twice (for qibolab 0.1 and 0.2).

I agree, we can postpone it (at least until we have the qibocal paper). Since we have three qubits we should still be able to play with two qubit gates.

PR #172, opened 2024-08-26
Update QM for new qubits/channels

alecandido:
https://github.com/qiboteam/qibolab/pull/1001

- [x] register QM configs https://github.com/qiboteam/qibolab/pull/995

### DIFF
--- a/qw11qD/platform.py
+++ b/qw11qD/platform.py
@@ -1,8 +1,8 @@
 import pathlib
 
 from qibolab.components import AcquireChannel, Channel, DcChannel, IqChannel
-from qibolab.instruments.qm import Octave, QmConfigs, QmController
 from qibolab.identifier import ChannelId
+from qibolab.instruments.qm import Octave, QmConfigs, QmController
 from qibolab.instruments.rohde_schwarz import SGS100A
 from qibolab.parameters import ConfigKinds
 from qibolab.platform import Platform
--- a/qw11qD/platform.py
+++ b/qw11qD/platform.py
@@ -1,5 +1,4 @@
 import pathlib
-from typing import cast
 
 from qibolab.components import AcquireChannel, Channel, DcChannel, IqChannel
 from qibolab.instruments.qm import Octave, QmConfigs, QmController
@@ -7,6 +6,7 @@ from qibolab.identifier import ChannelId
 from qibolab.instruments.rohde_schwarz import SGS100A
 from qibolab.parameters import ConfigKinds
 from qibolab.platform import Platform
+from qibolab.platform.platform import QubitMap
 from qibolab.qubits import Qubit
 
 FOLDER = pathlib.Path(__file__).parent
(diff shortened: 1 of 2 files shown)


qw11qD/platform.py
```diff
@@ -12,64 +15,72 @@
...
+        f"D{i}": Qubit(
+            drive=f"D{i}/drive",
+            flux=f"D{i}/flux",
+            probe=f"D{i}/probe",
+            acquisition=f"D{i}/acquisition",
         )
+        for i in range(1, 6)
+    }
 
     # Connect logical channels to instrument channels (ports)
```
stavros11, line 39: This comments seems outdated now, I would change to
```suggestion
    # Create channels and connect to instrument ports
```
or just drop completely.

qw11qD/platform.py
```diff
@@ -12,64 +15,72 @@
...
-            QmChannel(qubits["D1"].drive, "octave5", port=2),
-            QmChannel(qubits["D2"].drive, "octave5", port=4),
-            QmChannel(qubits["D3"].drive, "octave5", port=5),
-            QmChannel(qubits["D4"].drive, "octave6", port=5),
-            QmChannel(qubits["D5"].drive, "octave6", port=3),
-        ]
-    )
+    channels |= {
+        qubits["D1"].drive: IqChannel(
+            device="octave5", path="2", mixer=None, lo=lo_map["D1"]
```
stavros11, line 55: Since we are creating the drive channels one by one anyway and not in a loop, we could even drop `lo_map` and hardcode the LO names here. `lo_map` is not used anywhere else anymore.

qw11qD/platform.py
```diff
@@ -12,64 +15,72 @@
...
+    }
     # Flux
-    channels.extend(
-        QmChannel(qubits[f"D{q}"].flux, "con9", port=q + 2) for q in range(1, 6)
-    )
+    channels |= {
+        qubits[f"D{q}"].flux: DcChannel(device="con9", path=str(q + 2))
+        for q in range(1, 6)
+    }
+    channels = cast(dict[ChannelId, Channel], channels)
```
stavros11, line 75: Is this really needed? I have an idea what this does (even though I had to search `typing.cast`), but it may look a bit weird to other users.

stavros11 (commented):
Looks good to me, I'd just keep on hold until we confirm that it works well on hardware using the corresponding qibolab branches.

alecandido:
> Looks good to me, I'd just keep on hold until we confirm that it works well on hardware using the corresponding qibolab branches.

Yes, definitely.

I also wanted to test with Qibocal (I have my own branch, based on yours, but for a weird reason related to submodules is now living in a fork https://github.com/alecandido/qibocal/tree/qibolab0.2-qubits - I will bring it back as soon as I'm ready).

qw11qD/platform.py
alecandido: I guess that to most users, using the VS Code Python extension, it doesn't look that weird :P

In any case, it was mainly for me, since I was using the type checker to refactor (here we don't have tests, other than actually loading and using the platform), so I reduced the noise with some extra hints.

Maybe I can simply declare the type during the original channel definition (as an empty dict).

qw11qD/platform.py
alecandido: Done in 3ffa187

qw11qD/platform.py
alecandido: Done in c97da13

qw11qD/platform.py
alecandido: Replaced in c97da13

stavros11 (approved):

PR #175, opened 2024-08-28
Load QM custom configs in qw11qD 

alecandido:
I didn't test it explicitly yet, but with this PR it should be possible to also use https://github.com/qiboteam/qibolab/pull/995

### DIFF
--- a/qw11qD/parameters.json
+++ b/qw11qD/parameters.json
@@ -46,7 +46,7 @@
             "frequency": 7137520000
         },
         "D1/acquisition": {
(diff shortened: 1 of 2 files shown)


stavros11 (approved):
Thanks @alecandido. As said in https://github.com/qiboteam/qibolab/pull/995#pullrequestreview-2267281001 this works, so we can merge. 

In 77ac9784d604c3105d83017bfbe983361c9d979f I just simplified the import, since `QmConfigs` is available top level in `qm` so we don't need the `.components`.

PR #180, opened 2024-09-17
Recalibration of line B of qw11q

stavros11:
Qubit | T1 (us) | T2 (us) | Assignment Fidelity | Gate infidelity (1e-3) |
-- | -- | -- | -- | -- 
B2 | 20 | 12 | 97% | 2.00 (0.02)
B3 | 24 | 21 | 94% | 1.61 (0.02)
B5 | 34 | 28 | 95% | 0.61 (0.03)

* Single shot: http://login.qrccluster.com:9000/V4wiVyGyTy2D5y394Q4mXw==
* Rabi amplitude: http://login.qrccluster.com:9000/sTG9-BnIRVS1QxSJd_RPqQ==
* T1: http://login.qrccluster.com:9000/8S6hIAl4Q0OqD8dlzc00SA==
* T2 B2: http://login.qrccluster.com:9000/KMN4NjxMQP2byzlSKd6QDQ==
* T2 B3: http://login.qrccluster.com:9000/5U0NfVToQti_SW-gR87KKg==
* T2 B5: http://login.qrccluster.com:9000/rbmarqg4TAmNSaF_tdFtTw==
* RB B2: http://login.qrccluster.com:9000/Dutn72szTrabOQsCSzgWLA==
* RB B3: http://login.qrccluster.com:9000/skFVeQiTSIWKWFDKcJEC4g==
* RB B5: http://login.qrccluster.com:9000/pJijci52SZmBQ-683EQjFw==

PR #187, opened 2024-10-09
Platform specific Readme

BrunoLiegiBastonLiegi:
This PR adds a small script to parse the `parameters.json` file and generate a `README.md` with some basic information for each platform.
For instance, for `qw5q_platinum`:

# qw5q_platinum

## Native Gates
**Single Qubit**: RX, RX12, MZ

**Two Qubit**: CZ

## Topology
**Number of qubits**: 5

**Qubits**: [0, 1, 2, 3, 4]

```mermaid
graph TD;
    0 <--> 2;
    1 <--> 2;
    2 <--> 3;
    2 <--> 4;
```


alecandido:
I like the simplicity of using Mermaid, and letting it do whatever prefers.

However, I checked the Mermaid docs, and they do not have any actual graph drawing class, so for large chips it may be doing a terrible job... (which is not a big deal right now)

For this PR, just using Mermaid is more than fine.
However, we may consider leaving an issue open to explore alternatives like Graphviz
https://www.graphviz.org/gallery/
which could also be simply integrated passing through Networkx (e.g. just loading the pairs into a `networkx` class, and asking it to dump the Dot file).
Other solutions are listed here https://networkx.org/documentation/stable/reference/drawing.html

But, as I said, these are ideas for later on. Right now is fine as it is, and incredibly nice in it's minimal implementation. Thanks :)

BrunoLiegiBastonLiegi:
Yeah indeed for the 11q is not working great:

```mermaid
graph TD;
    A1 <--> A2;
    A1 <--> A3;
    A1 <--> D5;
    A2 <--> A4;
    A2 <--> A6;
    A3 <--> A4;
    A3 <--> D4;
    A4 <--> A5;
    A4 <--> B3;
    A5 <--> B1;
    A6 <--> B3;
    A6 <--> D3;
    B1 <--> B2;
    B1 <--> B3;
    B2 <--> B4;
    B3 <--> B4;
    B4 <--> B5;
    D1 <--> D2;
    D1 <--> D3;
    D2 <--> D4;
    D3 <--> D4;
    D4 <--> D5;
```

I tried to look for alternative layout algorithm but I couldn't find much information...

alecandido:
> Yeah indeed for the 11q is not working great:

Well, it's not ideal, but not even that bad.

It's a bit entangled, but you get all the information you need.
Maybe we can tweak it a bit, e.g. using smaller boxes around names - and in any case you can zoom and pan, which is convenient - but for even larger circuits it may be suboptimal.
In any case, this diagram could even be the default, and even replaced with a manually tuned one, which may better represent the chip features (e.g. putting them on a specific grid, and then drawing some connections - which is most likely what happens even physically on the chip, to some extent...).

alecandido:
@BrunoLiegiBastonLiegi I'd discourage you to go through the workflows' README generation, for two main reasons:
1. the generated README may be manually edited, and you don't want to overwrite those (nor attempting generating a diff to apply, which is definitely too complex...)
2. in order to make it work, the workflows should commit to the repo, which is not that great...

Instead, what you could do is to check in the workflow if any platform is lacking a README, and then publish a message on the PR suggesting providing such a README, and a good starting point may be achieved by running the script for that platform.

BrunoLiegiBastonLiegi:
Ok this should be more or less ready, the only thing I am not completely sure about concerns the native gates.
First of all, what gate is the so called `RX12` that you find in the `parameters.json` is it the `GPI2`? Secondly, the `RZ` gate is never explicitely reported in the configuration, but I believe we always have since it's just a virtual phase, should I always add it automatically (the same applies to the `Z` maybe)?

generate_readme.py
```diff
@@ -0,0 +1,117 @@
...
+    return readme_str
+
+
+if __name__ == "__main__":
+
+    platforms = [
+        directory.name
+        for directory in os.scandir(".")
+        if directory.is_dir()
+        and not (directory.name.startswith("_") or directory.name.startswith("."))
```
alecandido, line 95:
```suggestion
        for directory in Path().glob("[!._]*/")
```
importing `Path` from `pathlib` on top.

generate_readme.py
```diff
@@ -0,0 +1,117 @@
...
+    ]
+
+    for platform in platforms:
+
+        filename = f"{platform}/parameters.json"
```
alecandido, line 100: If `platform` is a `pathlib.Path`, much better to do:
```suggestion
        filename = platform / "parameters.json"
```

generate_readme.py
```diff
@@ -0,0 +1,117 @@
...
+        except RuntimeError as err:
+            warnings.warn(
+                err.args[0]
+                + f" Unable to generate the README for platform ``{platform}``."
+            )
+            continue
+
+        with open(f"{platform}/README.md", "w") as f:
+            f.write(f"# {platform}\n")
+            f.write(readme_str)
```
alecandido, line 117:
```suggestion
        (platform / "README.md").write_text(f"# {platform}\n{readme_str}")
```
to me looks simpler, but it's of course debatable.

alecandido (commented):
Ok, I didn't go through the whole script, and it's brief enough that is not worth optimizing that much...

But here are just a few technical and qualitative simplifications.

generate_readme.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+if __name__ == "__main__":
+
+    platforms = [
+        directory.name
+        for directory in os.scandir(".")
+        if directory.is_dir()
+        and not (directory.name.startswith("_") or directory.name.startswith("."))
+    ]
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: yeah it was original like that, just with `sys.argv` instead of `argparse`, but at some point I realized that it was less annoying (at least during testing) to just generate all the readme in one go without the need to manually specify.
But I can revert it back.

Edoardo-Pedicillo:
Could we use this PR to add the installation requirements in the READMEs https://github.com/qiboteam/qibocal/issues/988#issuecomment-2380572810 ?

alecandido:
> Could we use this PR to add the installation requirements in the READMEs [qiboteam/qibocal#988 (comment)](https://github.com/qiboteam/qibocal/issues/988#issuecomment-2380572810) ?

Manually yes (with the script is overly complicated, for little to no advantage...).

Or even a PR on top of this one is also fine.

generate_readme.py
alecandido: Uhm...

```py
[ins] In [2]: list(Path().glob("[!._]*/"))
Out[2]:
[PosixPath('qw11qD'),
 PosixPath('qw5q_platinum'),
 PosixPath('qw11q'),
 PosixPath('iqm5q')]
```

alecandido: Are you sure you didn't forget the final `/`?

alecandido: Also tested on macOS now:
```py
[ins] In [2]: list(Path().glob("[!._]*/"))
Out[2]:
[PosixPath('tests'),
 PosixPath('examples'),
 PosixPath('doc'),
 PosixPath('htmlcov'),
 PosixPath('src')]
                                                                                                                                                                                                                
[ins] In [3]: list(Path().glob("[!._]*"))
Out[3]:
[PosixPath('CODE_OF_CONDUCT.md'),
 PosixPath('flake.lock'),
 PosixPath('selfhosted'),
 PosixPath('LICENSE'),
 PosixPath('pyproject.toml'),
 PosixPath('tests'),
 PosixPath('tags'),
 PosixPath('README.md'),
 PosixPath('flake.nix'),
 PosixPath('examples'),
 PosixPath('doc'),
 PosixPath('coverage.xml'),
 PosixPath('poetry.lock'),
 PosixPath('htmlcov'),
 PosixPath('src')]
```

Which Python version are you using?

BrunoLiegiBastonLiegi: 3.10.15

alecandido: Found https://docs.python.org/3/library/pathlib.html#pathlib-pattern-language

> Changed in version 3.11: Globbing with a pattern that ends with a pathname components separator ([sep](https://docs.python.org/3/library/os.html#os.sep) or [altsep](https://docs.python.org/3/library/os.html#os.altsep)) returns only directories.

alecandido: The patch could be reverting to the previous behavior, for compatibility:
```py
        for directory in Path().glob("[!._]*/") if directory.is_dir()
````

File: .github/workflows/readme.yml [hunk 5]
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v5
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'If you are changing the configuration of one or more platforms remember to run `python generate_readme.py PATH_TO_PLATFORM` to update the `README.md` accordingly.'
```
alecandido, line 16:
```suggestion
              body: 'If you are changing the configuration of one or more platforms remember to run `python generate_readme.py path/to/platform ...` to update the `README.md` accordingly.'
```

alecandido (approved):
The last ones were just stylistic corrections. To me, it's good to merge.

.github/workflows/readme.yml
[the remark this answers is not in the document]
alecandido: #189

Edoardo-Pedicillo (approved):

scarrazza:
@BrunoLiegiBastonLiegi could you please fix the test issue?

BrunoLiegiBastonLiegi:
Not sure the error is related to this PR specifically. Even #180 was failing for the same reason.
By the way, @stavros11 @alecandido, is this not supposed to work with `0.2` due to the `parameters.json` being changed significantly? If that's the case, should I create the analogous script for `0.2`?

alecandido:
No, you're not supposed to fix that error. But the problem is not 0.2, but rather the OPX1000. The *stable* workflow is testing with the latest release, and the 0.1 driver for the OPX1000 is yet unreleased (in 0.1, while it has been just released for 0.2). Instead, the driver is merged in the `0.1` branch of Qibolab (the `main` equivalent for the 0.1 series), and that's what the *dev* workflow is pulling.

So, the workflows should be fine as they are.

alecandido:
> By the way, @stavros11 @alecandido, is this not supposed to work with `0.2` due to the `parameters.json` being changed significantly? If that's the case, should I create the analogous script for `0.2`?

You depend very mildly on the format of `parameter.json`, but still enough to break, I'm sorry...
https://github.com/qiboteam/qibolab_platforms_qrc/blob/be32893a50f6d7e4d778cfcc84e4238cd13cca0d/generate_readme.py#L21
none of these keys is any longer present.

Now, you should extract the `qubits` as the keys of the `native_gates.single_qubit` dictionary
https://github.com/qiboteam/qibolab_platforms_qrc/blob/b2f6b356dccc924bea4a841f954afe644f55b378/qw11q/parameters.json#L183-L185
while the `nqubits` will be just `len(qubits)`. And analogously with the `topology`, that is derived from the available two-qubit gates
https://github.com/qiboteam/qibolab_platforms_qrc/blob/b2f6b356dccc924bea4a841f954afe644f55b378/qw11q/parameters.json#L387-L388

BrunoLiegiBastonLiegi:
This is ready, but it's not going to be merged in main, rather in `0.1`.
I will create another PR to implement this for the current main.

PR #188, opened 2024-10-10
qw5q_platinum controlled with QM OPX1000 for qibolab 0.2

stavros11:
Requires https://github.com/qiboteam/qibolab/pull/1068. Updated qibocal routines have been tested to work: http://login.qrccluster.com:9000/f5mxayd_QXOwM9TcgpvEBw==.

### DIFF
--- a/qw5q_platinum/platform.py
+++ b/qw5q_platinum/platform.py
@@ -19,10 +19,7 @@ ConfigKinds.extend([QmConfigs])
 
 
 def create():
-    """Line D of QuantWare 21q-chip controlled with Quantum Machines.
-
-    Qubits D1, D2, D3 have been tested to work.
-    """
+    """QuantWare 5q-chip controlled with Quantum Machines OPX1000 and Octaves."""
     qubits = {i: Qubit.default(i) for i in range(5)}
 
     # Create channels and connect to instrument ports
--- a/qw5q_platinum/platform.py
+++ b/qw5q_platinum/platform.py
@@ -1,0 +1,77 @@
+import pathlib
+
+from qibolab import (
+    AcquisitionChannel,
+    Channel,
+    ConfigKinds,
+    DcChannel,
+    IqChannel,
+    Platform,
+    Qubit,
+)
+from qibolab.instruments.qm import Octave, QmConfigs, QmController
+from qibolab.instruments.rohde_schwarz import SGS100A
+
+FOLDER = pathlib.Path(__file__).parent
+
+# Register QM-specific configurations for parameters loading
+ConfigKinds.extend([QmConfigs])
+
+
+def create():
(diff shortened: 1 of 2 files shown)


qw5q_platinum/platform.py
```diff
@@ -0,0 +1,77 @@
...
+
+# Register QM-specific configurations for parameters loading
+ConfigKinds.extend([QmConfigs])
+
+
+def create():
+    """Line D of QuantWare 21q-chip controlled with Quantum Machines.
+
+    Qubits D1, D2, D3 have been tested to work.
+    """
```
alecandido, line 25: I guess this is a leftover from a copy...

stavros11: Fixed in 49467d3e1f02a5a377990fea6b8321ba3be8879c, thanks.

alecandido (approved):
I've seen the report, nice!

Since `main` will be now reserved for 0.2 (or at least `rm0.1` is), I guess we could even merge this (it's the only platform working with the platinum and 0.2)

stavros11:
> Since `main` will be now reserved for 0.2 (or at least `rm0.1` is), I guess we could even merge this (it's the only platform working with the platinum and 0.2)

qw11q also works with 0.2 using #153. They also both work with 0.1.

Generally, after this everything in the XLD1000 that is controlled with QM works with both 0.1 and 0.2. It just needs the right combination of branches and qm-qua versions.

alecandido:
Yeah, my point was that is the only one working for 0.2 AND the platinum. For this combination there is no alternative, so it should be fine to merge immediately into any 0.2 branch (either `rm0.1` or `main`, when it will be).

alecandido:
> qw11q also works with 0.2 using #153. They also both work with 0.1.

That's a good reason to merge even that PR

stavros11:
Yes, then if you agree we can merge all the 0.2 working platforms (qw11q and this) to `rm0.1`. It is just that the workflow will be failing until https://github.com/qiboteam/qibolab/pull/1068 is merged (for main workflow) and released (for stable workflow).

alecandido:
> Yes, then if you agree we can merge all the 0.2 working platforms (qw11q and this) to `rm0.1`. It is just that the workflow will be failing until [qiboteam/qibolab#1068](https://github.com/qiboteam/qibolab/pull/1068) is merged (for main workflow) and released (for stable workflow).

Yes, we are aware of it, but we're lacking a better alternative at the moment. So, better not to get stuck after a workflow (since we still have a report). For #153 they are even passing, so no problem at all.

Workflows are not useless, but they are there to check that things are not getting forgotten by chance.

So, I'd merge (but that's my opinion). In the meanwhile, I'll review https://github.com/qiboteam/qibolab/pull/1068 (but don't wait for that, the platform should not be affected almost at all, since the interface is mostly driven by Qibolab).

PR #192, opened 2024-10-23
Restore calibration parameters

andrea-pasquale:
Adding for the platform currently in `main` a `calibration.json` to store the calibration parameters.
See https://github.com/qiboteam/qibocal/pull/1021 for more information.


stavros11 (approved):
Even though still a draft, it will be needed if we merge https://github.com/qiboteam/qibocal/pull/1021 and it is not breaking anything (worst case `calibration.json` will be ignored), so it should be fine to merge.

Other than that, I am not sure if `calibration` and `parameters` are the best names - personally I would expect that `calibration` is something that is actually used during execution (by qibolab), which is not the case for us, but if it is documented it shouldn't be a big issue.

alecandido:
Yes, *calibration* is closer to the content of `parameters.json`, while the current `calibration.json` mostly contains characterization parameters (values that identify the specific physical, rather than the optimal ones to control it in a certain way).

However, it is true that it may even contain parameters which are neither (as validation ones, e.g. the RB fidelities), so it is harder to find it a perfectly suitable name (other than `parameters-leftover.json`). The idea of `calibration.json` is just to mirror the *cal* in Qibocal - we could call it `qibocal.json`, but it may be weird as well for other reasons.

If you have a fuzzier term than *calibration*, we could use that. We can even use `characterization.json`, as in general more accurate (though not completely) - but it is further from Qibocal and harder to type :P
We just need any random name, and to document it well. And giving meaningful names is [one of the hard problems in CS](https://martinfowler.com/bliki/TwoHardThings.html)... (and [science](https://en.wikipedia.org/wiki/Quark#Etymology) [in general](https://en.wikipedia.org/wiki/Axion#Prediction))

stavros11:
My initial idea was `parameters -> calibration` and `calibration -> characterization`, but this requires breaking qibolab 0.2 so it's not an option. In any case, I am fine leaving as it (following the qibo-cal argument) is or just renaming `calibration -> characterization`.

Another thing to note is that QM platforms may also have a `calibration_db.json` (autogenerated - not sure if I can change the name but I can check), which may add more confusion but it's not an exact conflict so it won't be an issue in practice.

andrea-pasquale:
Thanks @stavros11, indeed we should merge. I will double check if I forgot to push something since a few things were changed in https://github.com/qiboteam/qibocal/pull/1021 in the meanwhile. If it works we can merge.

alecandido (approved):

andrea-pasquale:
I thought that it might be a good idea to add here also channels for the 12 transition before merging this PR.
I've just pushed the updated version for the `qubit_spectroscopy_ef` protocol in https://github.com/qiboteam/qibocal/pull/1021. Here are the [results](https://github.com/qiboteam/qibocal/pull/1021).
It seems to work however, if I am not mistaken before we were measuring an anharmonicity of around 211 MHz now it is roughly 300 MHz.
@alecandido @stavros11 feel free to have a look to check that the way in which I added the new channels is correct.
If so, feel free to merge.

alecandido:
@andrea-pasquale I will check it better later on, but at first it seems good to me.

I'm still thinking of implementing an easy way to expand the platforms' channels during the experiment definition, since the information you're providing for the transitions is the exact same of the corresponding drive channel (since the physical medium is the same), and you should not be forced to repeat it.

Since the instrument object is still mutable, an easy way would be to do something like:
```py
class Controller(...):
    def add_transition(self, drive: ChannelId, transition: TransitionId):
        ...
```
But it may be even better to do it immutably
```py
class Controller(...):
    def with_transition(self, drive: ChannelId, transition: TransitionId) -> Self:
        ...
```
and then expose somehow in the execute method of the `Platform`, e.g.
```py
class ExecutionParameters:
    ...
    transitions: list[TransitionId]
```

The only pain point is that you will need to calibrate natives assigning them channels which do not exist in the original platform. We could try to do something to improve do situation, like:
```py
class ExecutionParameters:
    ...
    transitions: Optional[list[TransitionId]] = None
```
and in the case in which is `None`, we try to infer the extra channels from the natives (but we will need to extend the `Pulse` to have a `TransitionId`, and do something convoluted, since we should look at the `RX` to check which is the drive channel, since in principle we do not know even in which instrument is the physical channel - everything quite messy).

This is just a rough sketch, but if you believe it may be useful to have something of this, or keep discussing the thing, we could open an issue in Qibolab to continue the design process.
If you believe the current setup is just good, we can leave it as it is.

My perspective is that it is actually useful to play at the level of logical channels in the sequences, since it makes it closer to the actual execution. But the physical part should remain confined to the initial platform layout, we should not need to be modified because further transitions are calibrated (you should not need to find which is the relevant Octave to calibrate a new transition). So, we may be missing a separate physical layer abstraction... @stavros11

andrea-pasquale:
> Both the channel addition and the routine update look correct to me. Btw, I think the link under "results" is not correct, it is linked to a PR instead of the actual results (I was expecting a report).

You are right, here is the correct link http://login.qrccluster.com:9000/_lJj3FqjSSuHY5mRd7C1hg==/

> I am not sure what is causing the difference in anharmonicity. If we want to exclude potential issue with 0.2, I would repeat the same measurement, on the same qubits, with the same calibration, using 0.1. If the results agree then most likely it's not a qibolab issue, otherwise I would check the QUA scripts to find the difference. The first part (reproducing and comparing 0.1 and 0.2 results) is quite tedious especially without an automated way of converting platforms, so if we want to go to that direction in may be worth investing time in the conversion script. A quicker check would be to just generate and compare the QUA scripts without worrying about actual executions.

I was doing something wrong with the anharmonicity, I fixed the problem in https://github.com/qiboteam/qibocal/pull/1036/commits/b094c4eb1ed53530641eb9a97c67863348e60750
 
> In any case, I would probably move this discussion to a qibolab issue, as I would expect this PR to be merged soon, since the corresponding qibocal PR is already merge (to `0.2` branch). I agree that it might be worth retesting on hardware before merging (I will also try to do that), as many changes happened since I last tried.

I agree, from my side we can merge.

alecandido:
> I agree, from my side we can merge.

Even from my side, the proposal to improve was not to be implemented right now. I'll move the discussion into an issue and tag both of you :)

PR #161, opened 2024-08-05
CZs and iSWAPs characterized

DavidSarlle:
All the iSWAPs and CZs available (coupler 0 not working so q0-q2 2q gate can not be characterized) characterized. 

This runcard need the implementation of qibolab issues (to run with qibolab/qibocal main): 

https://github.com/qiboteam/qibolab/issues/921
https://github.com/qiboteam/qibolab/issues/920
https://github.com/qiboteam/qibolab/issues/900

Fine tunning of the CZs  landscape also need the implementation of the issues 921, 920 and 900 is needed to be able to run landscape without fitting errors (already fixed in qibocal main)

PR #196, opened 2024-11-03
Qblox: iqm5q

alecandido:
Attempt to port https://github.com/qiboteam/qibolab_platforms_qrc/tree/david/iqm5q_CZs_iSWAPS_refined/iqm5q for the new Qblox driver in Qibolab 0.2.

- [ ] parameters need to be imported from #161  
    - they are currently copied from `qw5q_platinum` ones


alecandido:
Also ported the most recent calibration from #232, but I did not manage to reproduce the results with 0.2 (exactly with https://github.com/qiboteam/qibolab/commit/939ea04470ff2c7ab812381b2f9a059eba1c99e0)

http://login.qrccluster.com:9000/qzzmNBKMTLCgZN3qNqnn_A==/

alecandido:
No need any longer, since #238 has been merged.

stavros11:
> No need any longer, since #238 has been merged (relying on the new Qblox driver).

Indeed. Just to note that this also contains the `qblox-test` platform which is just a loopback and not connected to any qubit, but is still available and could be useful for some debugging (if a qubit is not needed) while the other platforms are not free. Not sure if we should have it on main though.

PR #199, opened 2024-11-06
New platform tii3q98 zcu111 for testing software

JavierSerranoGarcia:
Platform and runcard for tii3q98 chip qubits 1, 2, 3 controlled with ZCU111 with a basic characterization. This PR works with 

qibo                      0.2.12
qibocal                   0.1.1      
qibolab                   0.1.10     
qibosoq                   0.1.4       
qick                      0.2.249

The platform has the usual issue in this board: the polyphase filter the Qick uses in this firmware does not allow demultiplexing the 3 readouts of the qubits at the same time. I suggested using qubit 1 or qubits 2,3.

@alecandido , is this what you need to test qibosoq?


alecandido:
Hi @JavierSerranoGarcia, sorry for the late reply.

Yes, that's perfectly fine!
In practice, even a single qubit should be sufficient to test all the range of features, since two-qubit gates are not using any new line or feature (other than possibly the pre-distortions), so we should not need to use more than one just to test most of the code, and two only for multiplexing itself (otherwise not even that).

Btw, the PR is based on some old commit, polluting the diff with apparent removals. May I clean it up a bit, without touching `tii3q98_zcu111/` at all?

JavierSerranoGarcia:
@alecandido, please remove everything you think should disappear.

alecandido:
I just rebased this branch on `0.1`, just to be ahead of that, and remove the apparent platforms' deletions (which were already removed in the updated `0.1` branch).

PR #200, opened 2024-11-07
calibrate mixers line B

Edoardo-Pedicillo:
This PR recalibrates the readout mixers of line B, and this solves the ring pattern showed by @hay-k (http://login.qrccluster.com:9000/hNe0JPXhQHmSuWRL6GPAWg==/). Considering that:
- the B1 readout frequency is 7147636454 Hz
-  the B4 readout frequency is 7712222447 Hz
-  the LO frequency is 743000000 Hz,

it follows that the IF frequencies are around +-282 MHz, and this could affects the results  if the mixers are not calibrated. (thanks @stavros11 for the calibration script).


stavros11:
Thanks for adding the mixer calibration @Edoardo-Pedicillo. I would suggest pushing also the updated platform parameters in this PR, because if I just use it for the single shot out of the box I still get the rings: http://login.qrccluster.com:9000/4gQq9kMFQ5K31qLkLG768g==
Instead, if use the parameters from your report (http://login.qrccluster.com:9000/hNe0JPXhQHmSuWRL6GPAWg==/new_platform/parameters.json), the problem is solved (http://login.qrccluster.com:9000/yjQrBitVTVymczADpuzwYg==), I guess because you also ran the other calibration routines as well.

A note regarding the mixer calibration script: If you tried to use it several times, you may have noticed that is not very stable, sometimes it gives errors and produces calibration with nan values. Long time ago QM told me that some of these issues were fixed in qm-qua 1.1.7, but we need to upgrade the cluster's firmware in order to use this version (currently we are stuck at 1.1.6). I may try to upgrade in the next downtime, but since it is not particularly urgent I would avoid doing it while the chips are operative.

hay-k:
@Edoardo-Pedicillo et al. thanks for figuring this out.

The fact that IF for B1 and B2 are so nicely symmetrically placed on the two sides of the LO (hence causing the image of one interfere with the other) is just a really bad choice of LO frequency. I don't think anyone has put much thought behind the chosen LO frequency (given that B line was not even fully calibrated in the past, so the exact RO frequencies were never known until we calibrated it recently), and I don't think it is special, so it can be changed. It is true that mixer calibration mitigates this issue, but I think a more robust solution is just to change the LO frequency to a better value based on the knowledge we have gained so far. To be clear, I am not saying mixer should not be calibrated - it should be for sure, however I think we should not rely on mixer calibration as a solution of this problem. This way we can also increase the readout fidelity of B4 as well, and the solution would be more stable, and independent of small changes in RO frequencies.

Slightly changing the LO frequency completely messes up things though - the resonance frequency jumps somewhere else, and I believe @alecandido mentioned that you have observed a similar thing. This really shouldn't happen, and I think it is better to try to investigate this issue, and figure out if it is driver issue, instrument/firmware issue, or data post processing issue.

stavros11:
> Slightly changing the LO frequency completely messes up things though - the resonance frequency jumps somewhere else, and I believe @alecandido mentioned that you have observed a similar thing. This really shouldn't happen, and I think it is better to try to investigate this issue, and figure out if it is driver issue, instrument/firmware issue, or data post processing issue.

We already discussed this with @hay-k, just writing here for others interested (@alecandido @Edoardo-Pedicillo): The issue is that the LO frequency appears in two places in parameters.json, once for the output: https://github.com/qiboteam/qibolab_platforms_qrc/blob/4219c5731799c97dde86198d8f3f9acb16cd01b6/qw11q/parameters.json#L204-L207
and once for the input: https://github.com/qiboteam/qibolab_platforms_qrc/blob/4219c5731799c97dde86198d8f3f9acb16cd01b6/qw11q/parameters.json#L224-L226
If you change the LO frequency in only one of those places, things are messed up. If both are changed, resonators come back to their original frequencies (as expected) and the ring issue is also solved, even without mixer calibration.

It is weird that the instrument allows us to use different frequencies for up and down conversion. I am not sure if that has any practical application. In principle, I could restrict this in the driver. However, since it is already fixed in 0.2, I would just leave it as it is and ask people to update both places when changing LO frequency.

To add to this, as mentioned before and as shown in the Octave schematic here: https://docs.quantum-machines.co/latest/docs/Hardware/octave/?h=octave#octave-block-diagram
note that the Octave has five outputs but only three LOs. Outputs 2, 3 and 4, 5 share LOs. Again, their API and our parameters.json allow using different frequencies for those, but one of them will actually be ignored!

Edoardo-Pedicillo:
> The fact that IF for B1 and B2 are so nicely symmetrically placed on the two sides of the LO (hence causing the image of one interfere with the other) is just a really bad choice of LO frequency. I don't think anyone has put much thought behind the chosen LO frequency (given that B line was not even fully calibrated in the past, so the exact RO frequencies were never known until we calibrated it recently), and I don't think it is special, so it can be changed. It is true that mixer calibration mitigates this issue, but I think a more robust solution is just to change the LO frequency to a better value based on the knowledge we have gained so far. To be clear, I am not saying mixer should not be calibrated - it should be for sure, however I think we should not rely on mixer calibration as a solution of this problem. This way we can also increase the readout fidelity of B4 as well, and the solution would be more stable, and independent of small changes in RO frequencies.

I agree, I have recalibrated only the mixers in this PR to ensure this is the issue and also I am not sure which value to set.

In https://github.com/qiboteam/qibolab_platforms_qrc/pull/200/commits/77531bbe428557ff35a7f5a35c4ef3d2d308b1ef I have moved the LO frequency 100 MHz and recalibrated the platform http://login.qrccluster.com:9000/dAtdb7X2T-6QrhhdKfVOSQ==/ following @stavros11 comment, and the resonator frequencies are not shifted.

PR #203, opened 2024-11-15
Script to generate calibration.json

andrea-pasquale:
I've added a script `generate_calibration.py` which automatically generates the new `calibration.json` file, from the `parameters.json` compatibile with qibolab `0.2`.

The syntax is the same as the `convert.py` command from #201.
```
python convert.py qw11q/parameters.json iqm5q/parameters.json ...
```

and a new file `calibration_new.json` will be generated. I've added `_new` to avoid conflicts with my existing `calibration.json` files.

@alecandido if you want I could try to merge the two scripts to speed up the conversion.


alecandido:
> @alecandido if you want I could try to merge the two scripts to speed up the conversion.

Fine by me, keep going!

PR #207, opened 2024-12-20
Switch line D readout from octave5 to octave6

stavros11:
Due to issues with doing acquisition on QM `con6`, we decided to move the readout of qw11q line D to octave6 which is connected to `con8`. I have tried to run a few routines and they seem to work, however thorough recalibration is probably needed.

Here is the high-power resonator spectroscopy: http://login.qrccluster.com:9000/pIVIdAmJS9W3bgbvK6NUvw==


alecandido:
> Here is the high-power resonator spectroscopy: http://login.qrccluster.com:9000/pIVIdAmJS9W3bgbvK6NUvw==

When fits are failing because it is clearly not a Lorentzian, you may try to use the `fit_function: "s21"` parameter, as described in https://qibo.science/qibocal/stable/protocols/resonator_spectroscopy/resonator_spectroscopy.html#id5

Most likely you do not care that much, but it may be able to find automatically the value even when the bell shape is significantly distorted.

hay-k (approved):

biankawolo (approved):

PR #216, opened 2025-01-24
Updated parameters for qw5q , mixer calibration script (copied and mo…

Luca-Ben-Herrmann:
Updated parameters.json for the qw5q chip with good classification (http://login.qrccluster.com:9000/EwNJsl0xRoSEVdpjyav0fQ==).

Single qubit calibrated well, two qubit gate roughly calibrated but due to fluctuations not reliable. 

Added a mixer calibration script that was copied from the qw11q folder and modified accordingly to work with the wq5q.


stavros11 (approved):
Thanks @Luca-Ben-Herrmann. I guess some recalibration is now needed as the single shot classification is not as good as in your original post: http://login.qrccluster.com:9000/d_bGWZu6QaSdvqlg9N0i-g==. Maybe the sweetspots have shifted from your recent experiments.

We could still merge this and add the recalibration on top, given that you are also adding the mixer calibration scripts here.

Luca-Ben-Herrmann:
@stavros11 Yes, some of the sweetspots changed since I did this pr. As you suggested, I would still like to merge it since it includes the mixer calibration script as well as the updated calibration.db file.

hay-k (approved):

PR #219, opened 2025-02-17
Update qw5q_platinum cluster

stavros11:
These changes (together with qiboteam/qibolab#1149) are probably needed to run using the new cluster, however I am not getting any signal so I would keep as draft for now.


stavros11:
I ported the latest calibration from #221 here. There is some weird inconsistency with 0.1 in the single shot only for qubit 0: http://login.qrccluster.com:9000/Ewq7cY6VTNqfZ4S8aqbQJQ==. I am not yet sure what is causing this.

PR #234, opened 2025-06-10
Add missing dependencies in CI

stavros11:
After the refactoring of dependencies in qibolab.
Note that this will only work for QM, as we have not created the qibolab-* repositories for other drivers. I also had to install `qcodes` manually, as it is needed to instantiate the R&S LOs used as TWPA pumps. I am not sure if there is a more elegant solution,

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,6 +21,7 @@ jobs:
           python -m venv testenv
           source testenv/bin/activate
           pip install pytest
+          pip install qcodes
           pip install qibolab-qm
       - name: Install and configure poetry
         if: ${{ matrix.version == 'main' }}
@@ -32,6 +33,10 @@ jobs:
         run: |
           python -m venv testenv
           source testenv/bin/activate
+          git clone https://github.com/qiboteam/qibolab
+          cd qibolab
+          poetry install --no-interaction --with tests --all-extras
+          cd ..
           git clone https://github.com/qiboteam/qibolab-qm
           cd qibolab-qm
           poetry install --no-interaction --with tests --all-extras
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,6 +38,7 @@ jobs:
             cd $lib
             poetry install --no-interaction --with tests --all-extras
             cd ..
+          done
       - name: Test platforms
         run: |
           source testenv/bin/activate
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -35,6 +35,14 @@ jobs:
...
           cd qibolab
           poetry install --no-interaction --with tests --all-extras
+          cd ..
+          git clone https://github.com/qiboteam/qibolab-qm
+          cd qibolab-qm
+          poetry install --no-interaction --with tests --all-extras
+          cd ..
+          git clone https://github.com/qiboteam/qibolab-qblox
+          cd qibolab-qblox
+          poetry install --no-interaction --with tests --all-extras
```
alecandido, line 45: Given the standardized structure, maybe we can replace this with a bash `for` loop.

alecandido (approved):
Just one refactoring proposal, but other than that is certainly fine :)

.github/workflows/tests.yml
stavros11: bb8bb477e855c241e3c811d8d622454f3055478c, 9e7690146321878c5a6848734428ba80ed13e763

andrea-pasquale:
> I also had to install `qcodes` manually, as it is needed to instantiate the R&S LOs used as TWPA pumps. I am not sure if there is a more elegant solution

Perhaps just to make sure that the `qcodes` version installed is the correct one we could introduce an extra in qibolab (or have a separated package like qblox and qm). Currently, although the dependency is available in qibolab, it is just a poetry group which can be installed only through poetry and not through pip.

andrea-pasquale:
> @andrea-pasquale
> 
> > Perhaps just to make sure that the `qcodes` version installed is the correct one we could introduce an extra in qibolab (or have a separated package like qblox and qm).
> 
> At some point, it will be for sure a package. For the time being, we thought we could survive with whatever version of Qcodes (we should not be sensitive to that, unlike `qm-qua` and `qblox-instruments` - also because Qcodes is much more stable).
> 
> Extra are "deprecated" by now. It was the old solution for the instruments' dependencies, and we know by now all the limitations of this solution. So, we are not going to use it any longer.
> 
> > Currently, although the dependency is available in qibolab, it is just a poetry group which can be installed only through poetry and not through pip.
> 
> That's a temporary issue. It is in a Poetry group for the tests. But Qibolab's tests should be independent of driver dependencies. They will be, and the dependency will be removed from the group as well.
> 
> [qiboteam/qibolab#1234](https://github.com/qiboteam/qibolab/issues/1234)

That's fine my recommendation is just to document somewhere now that `qcodes` should be installed separately.
Also the usage of the new packages `qibolab-qm` and `qibolab-qblox` should be documented accordingly.

PR #240, opened 2025-08-21
Qblox loopback platform

alecandido:
https://github.com/qiboteam/qibolab_platforms_qrc/pull/196#issuecomment-3210281940

### DIFF
--- a/qblox-test/calibration.json
+++ b/qblox-test/calibration.json
@@ -1,0 +1,24 @@
+{
+  "single_qubits": {
+    "0": {
+      "resonator": {
+        "bare_frequency": 5221992000,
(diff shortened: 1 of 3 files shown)


stavros11:
@alecandido fyi, I have updated the firmware of this cluster to 0.11.0 (from 0.7.0), so that it agrees with the other clusters in the lab and can be used with the same `qblox-instruments` version (v0.16.0).

alecandido:
> @alecandido fyi, I have updated the firmware of this cluster to 0.11.0 (from 0.7.0), so that it agrees with the other clusters in the lab and can be used with the same `qblox-instruments` version (v0.16.0).

Thanks! 🙏

PR #241, opened 2025-08-22
[WIP] qw5q_platinum controlled with Qblox

stavros11:
Not getting nice results with qibocal yet, investigating the issues.


stavros11:
I was able to fix some of the issues with the single shot routine by using the correct acquisition durations, which in this case are larger than the corresponding probes. This requires a minor fix to the Qblox driver implemented in https://github.com/qiboteam/qibolab/pull/1278. 

This improved the results on some of the qubits, however qubits 2 and 3 are still not ideal (http://login.qrccluster.com:9000/_uoaNsoURFu8QLVGCJhqEg==). I am not yet sure why.

PR #249, opened 2025-10-28
qw21q line D

alecandido:
- offset scaled between -1 and 1


alecandido:
The pi-pulses are all decent, as you can see Rabi oscillations very clearly in the quadratures, and at least decently in classified shots.

However, the single shot classification itself is terrible:
- one single qubit classifies around 90% assignment fidelity
- another two of them classifies around 75-80%
- the last two are just around 60% (practically random...)

Coherences are also decent, with T1 ranging `~ 10-20 us`, and T2 at least `> 5 us`.

No result is great, and a TWPA recalibration may be needed (less relevant, but a mixer calibration would be also appreciated - just not possible with the current driver...).

However, the main goal of this calibration is to work as a playground for the Qblox driver development (based on `qblox-instruments`).
In this context, it is particularly relevant that one of the present goals was to test the filter's application. But even the mandatory preparatory step, `flux_amplitude_frequency`, is not giving any meaningful result. Specifically, the last batch of attempts are just showing completely flat curves, yielding 0 detuning for any flux pulse amplitude (while the qubit flux dependency, sweeping bias, is actually working as expected).

PR #257, opened 2025-12-04
Remove outdated loopback scheme platform

alecandido:

PR #258, opened 2025-12-05
run pre-commit on autogenerated READMEs 

RoyStegeman:
As suggested here https://github.com/qiboteam/qibolab_platforms_qrc/pull/257#issuecomment-3612874498

@alecandido also please have a look at the reading of the gate infidelity. The values of `rb_fidelity` are 0.0 for all the current `calibration.json` files so I can't check if the result (and thus processing) is sensible


generate_readme.py
alecandido: I would guess the complement should be wrt to 1: both fidelities and infidelities are probability-like.

```suggestion
        gate_infidelity_cv = 1 -qinfo["rb_fidelity"][0]
```

alecandido (approved):

generate_readme.py
RoyStegeman: Yes, I figured I had probably misheard you since this makes a lot more sense

PR #261, opened 2025-12-12
Lballerio calibrations

lballerio:
New calibration for qw21q-b chip, with calibrated B1-B3 2 qubit gate and single qubit gates for every qubit.


alecandido:
@lballerio since this branch is aging (and the calibration in `main` even more), I'm merging this. Whenever you want to update anything, just open a new one.

PR #287, opened 2026-04-07
remove bounds from parameters.json for qibolab 0.2.13 and remove iqm5q platfrom

RoyStegeman:
Since qibolab 0.2.13 the handling of the memory capacity of the electronics has been moved to the driver-level and thus including it in `parameters.json`, as was done with the `bounds` config, is no longer supported. This issue was raised in  https://github.com/qiboteam/qibolab_platforms_qrc/issues/288

Since the iqm5q chip is no longer connected to control electronics, the corresponding platform is removed from this repo.


scarrazza:
Did you test this for circuit execution?

RoyStegeman:
Yes I tested with 

``` python
from qibo import models, gates
import numpy as np
from qibo import set_backend


use_qpu = True
if use_qpu == True:
    set_backend("qibolab", platform="qw5q_platinum")
else:
    set_backend("numpy")

nqubits = 5
c = models.Circuit(nqubits)

# Define a list of angles (in radians)
# This creates a gradient from "mostly 0" to "mostly 1"
angles = [np.pi/8, np.pi/4, np.pi/2, 3*np.pi/4, 7*np.pi/8]

for i in range(nqubits):
    # RY rotates the vector toward the |1> state
    c.add(gates.RY(i, theta=angles[i]))

# Add measurements
c.add(gates.M(*range(nqubits)))

result = c(nshots=100)

print(result.frequencies())
```

which gives 

```
Counter({'00011': 31, '00111': 25, '00001': 9, '01011': 9, '00101': 7, '01111': 5, '10011': 3, '00110': 2, '11111': 2, '00000': 1, '00010': 1, '00100': 1, '01101': 1, '10001': 1, '10010': 1, '10111': 1})
```

So at least it's running.

scarrazza:
Thanks.

